PR #1, opened 2015-04-13
Search navigation fix

Scarfmonster:
The menus didn't expand properly


Scarfmonster:
I only just noticed someone reported that here http://mantis.tokeek.de/view.php?id=571

PR #2, opened 2015-04-19
English Synonyms and small fixes

Scarfmonster:
The English synonym file is somewhat huge but it's the only freely licensed English thesaurus I currently know of.


Orbiter:
great work! Thank you!

PR #3, opened 2015-04-20
How about normal font weight in the searched titles & fix RSS icon position?

shaman:


Orbiter:
good stuff, thank you

shaman:
I'm very glad to participate
Regards!

PR #5, opened 2015-05-06
Updated linux scripts so they can be started from any working directory

arucard21:
I've changed the Linux shell scripts so they no longer rely on being in a specific working directory in order to run correctly. They will automatically use the directory that contains the script as program directory, but this is a variable in the script which you can change to anything you like. 

This makes it easier to make start and stop YaCy. It should also make it easier to use with init and autostart systems.


arucard21:
I missed some parts of the startup script (it was changing into the program directory, which is why it worked). I fixed it so it now truly works without needing the program directory to be the current working directory. This exposed some bugs in YaCy itself though. 

While the startup script now works correctly and can start YaCy, once YaCy actually starts up, it doesn't seem to be able to find all files. This seems to be because new File objects aren't always created with `getDataPath()` or `getAppPath()`, e.g. in `net.yacy.search.Switchboard.java` on lines 770 and 854. This means they are created relative to the current working directory, which in this case will no longer be the program directory, so they can't be found.

sudheesh001:
@arucard21 Could you please squash the commits into one single one. Looks like this is working as of now. :smile:

arucard21:
I think I did it now. It's the first time I've squashed commits but it seems like all the changes are still correctly shown in the PR.

Orbiter:
I don't see the point to make these changes. Please give an example what did not work before and is working with your change.

arucard21:
The scripts used to "cd" into the program directory in order to run yacy with the necessary files and folders that are stored relative to the program directory. This could be problematic for other scripts using the yacy scripts since their working directory would suddenly be changed to the program directory. So I changed the scripts to explicitly define the paths to the necessary files and folders and provide them as command-line parameters, which were already available for yacy. Unfortunately, since they were never really used, yacy didn't actually work correctly with those command-line parameters, which is what @reger24 fixed in the commit that references this PR. 

An example from how it worked before this PR:
If your yacy program dir is in /usr/lib/yacy/ and you open your shell at /home/user/ and start the startup script by runing /usr/lib/yacy/startYACY.sh, it would start yacy but afterwards your shell's working directory wouldn't be /home/user/ anymore, but /usr/lib/yacy. 

Now if you want to start yacy by having your init system run /etc/init.d/startYACY.sh, it could be quite problematic for all the other programs started by the init system if your working directory suddenly changed. This might still start yacy, but it could break any other program starting after it. 

With this PR, the working directory never changes when you call yacy's scripts, so everything should work correctly.

arucard21:
Closed because stale

PR #9, opened 2015-07-09
Initial support for !bangs

cloutier:
This is the beginning of my work on implementing [DuckDuckGo's !bangs](https://duckduckgo.com/bang). I want a bit of feedback before I do the rest. I am new to java and very new to yacy. Please tell me if there is anything that can be improved, I will happily do it! :)

### DIFF
--- a/htroot/yacysearch.html
+++ b/htroot/yacysearch.html
@@ -29,6 +29,10 @@
     }
   }
   document.onkeydown = handleArrowKeys;
+  redirect = "#[redirection]#";
+  if (redirect != "") {
+  	window.location.href = redirect;
+  }
   </script>
   <script type="text/javascript" src="/env/bootstrap/js/typeahead.jquery.min.js"></script>
   <script type="text/javascript">
--- a/htroot/yacysearch.java
+++ b/htroot/yacysearch.java
@@ -36,6 +36,7 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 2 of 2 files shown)


cloutier:
@Orbiter I saw you were the last to commit. Do you know who does the code review?

reger24:
My opinion:
the !bang Feature is a bit out of Focus for yacy, if I understand !bang right, it Forwards a search just to a other search engine. This seems not the ideal interoperation for YaCy with other Systems.
YaCy supports several imports and opensearch to incorporate other Systems in the result list (even if there are not so many which actively Support this).
So far YaCy offers a "Compare Search" which Displays other search engines results. If you implement !bang in that "compare" context it might be nice, but a implementation which instead of YaCy results only Displays other engines results, without improving YaCy's index or results seems not befefitial to me.
In the context of opensearch-connectors it would be easy to implement a Kind of "Spy.Connector" which would improve YaCy index/results but I think that's not appropriate for a "Open Search Engine" like YaCy.

cu

cloutier:
It's a feature I really liked when using DuckDuckGo, because it allowed me to quickly use Google for the query it couldn't handle. I could change my default search engine without regretting it because of that. It is not designed to take the results of other services, but to augment YaCy's flexibility. 

I am working on making this a feature for power users, not enabled by default.

The idea of comparating is interesting, I'll look into that.

Scarfmonster:
From what I've seen the bangs to work as, it mostly looks like redirects to other search engines, or something out of scope of what yacy currently can do (or can do with site: switch) 

As for redirects, it may be interesting to allow users to define their own bangs in a syntax similar to how the current OpenSearch import works. Hardcoding them is not a good idea.

PR #10, opened 2015-07-21
Fixed CSS scrolling

Raegdan:
When the sidebar on search page becomes scrollable, the scrollbar shrinks the sidebar and makes the search results weirdly scrollable on X axis by several pixels. Now the sidebar always have a scrollbar, and results are never X-scrollable.

### DIFF
--- a/htroot/env/bootstrap-base.css
+++ b/htroot/env/bootstrap-base.css
@@ -35,7 +35,7 @@ body {
     display: block;
     padding: 20px;
     overflow-x: hidden;
-    overflow-y: auto; /* Scrollable contents if viewport is shorter than content. */
+    overflow-y: scroll;
   }
 }
 
@@ -123,4 +123,7 @@ ul.nav li.dropdown:hover ul.dropdown-menu{
 .popover {
     z-index: 1600;
 }
- 
+.col-md-9 {
+    width: auto;
+    overflow-x: hidden;
+}


Raegdan:
@Orbiter, what can you say about my patch after 2 months?
Accept it or tell what's wrong and I'll be glad to improve, but don't ignore! You state on your site you are looking for design improvements, right? ;)

reger24:
Just my oppinion:
I like eliminated horizontal x-scoll (by a couple pixel)
but not the permanent sidebar scroll bar.
So I'd keep line 38 (    overflow-y: scroll; /* Scrollable contents if viewport is shorter than content. */) as is and take the other modification.

Raegdan:
@reger24 as far as understood this problem, the sidebar scrollbar appearing/disappearing is the cause of search results shifting right and left respectively. So, if we make the scrollbar visibility dynamic, we will get rid of search results X-scrolling, but it will still shift by X axis during sidebar operations like expanding/collapsing the long widgets. May be there is a way to make the sidebar become transparent with geometry preserving instead of hiding it completely? Alas i'm not a CSS guru despite of tons of books I've read :(

reger24:
@Raegdan not on my browser (Firefox 40, IE11, Edge) there search stay fixed and navbar shrinks in x by width of scollbar on activation

PR #12, opened 2015-09-04
Updated french locale and added new translator utils

luccioman:
Continued updating french locale for non administrative pages.
Added utils  to launch translation of all source files without launching YaCy full application and to list non translated files.
Corrected minor errors on 2 source files.
Added javadoc and reusable FileNameFilter


Orbiter:
Great stuff, thank you!

PR #18, opened 2015-10-15
Add TIFF imge format support

luccioman:
Addef TIFF image format support by integrating haraldk/TwelveMonkeys ImageIO library :
- integrated twelvemonkeys jar files
- modified ImageParser.java to read images using ImageIO available plugins
- refactored ViewImage.java to ease understanding and performance mesurements
- created and used performance test class and file formats support HTML test file
- validated using YaCy image search filtering on '.tif' filetype


Orbiter:
Thank you for this work, I looked over all your changes in the code and found that is was made nicely and with great care. I tested the behaviour regarding the ability to find and show tiff images in the preview and it worked fine. However, there is a problem showing the enlarged preview but I did not yet found out if this is problem with the css/js framework which does the preview rendering.
However, thank you very much, thats a good work we can built on more image data format recognition.

PR #21, opened 2015-10-23
Allow full size preview of TIFF images and fixed some rendering issues

luccioman:
- made it possible to view TIFF images in full size preview even if browser does'nt support it, using ViewImage servlet instead of original source
- fixed some image preview rendering issues :
 - aspect ratio of gif and svg images
 - avoid 'Loading...' staying frozen when an image render error occured (unsupported format, 404 error etc..)
- added a test class (ViewImageTest) to allow testing ViewImage rendering on a folder containing multiple images


Orbiter:
great, thank you!

PR #26, opened 2015-11-20
Images rendering improvements

luccioman:
Improvements for images rendering in YaCy Image Search results :
- corrected default favicon management
- process image files as streams instead of complete byte arrays. Advantages :
   - image content has not to be fully loaded before detecting that format is not supported
   - temporary byte buffer is no more needed for gif or svg files : content stream is directly copied to HTTP response
   - image parsing has not to wait full content is loaded to begin
   - more control is available to eventually skip useless bytes
   - small image files are still cached even if processed as streams
- integrated haraldk/TwelveMonkeys imageio-bmp-3.2.jar library :
  - better support for differents flavors of BMP format
  - support transparency
  - support PNG encoded ICO image files
  - tested render with http://entropymine.com/jason/bmpsuite/ and http://bmptestsuite.sourceforge.net/ test suites
  - deprecated icoParser and bmpParser classes. They might be deleted.
- corrected ViewImage rescale operation
- removed a useless encoding operation at the end of ViewImage render

All tests run on : 
- OS : GNU/Linux Debian 8.2 (Jessie) 64 bits
- browsers :
 - Iceweasel 38.3.0
 - Konqueror 4.14.2
 - Chromium 45.0


Orbiter:
Hi Luc, thank you again for your work. Unfortunately I was involved in loklak for the last weeks so much that I did not find time to have a closer look at your work. I am still trying to find all new features you put in. I believe the image load can now do much more, but also produces much more exception logging. Can you have a look which kind of exceptions can be handled (i.e. encoding issue) and which one cannot (i.e. network errors). The encoding exceptions can be logged to have an idea what can be done. The network errors cannot be fixed, so just silently go over them.

luccioman:
Hi, thank you for your confidance and feedback.
I really did not add so much new features in my last work, but rather 
focused on improving image rendering performance for scaled and unscaled 
images, and on cleaning code to be as most generic as possible (see 
previous default favicon specific code and ico/bmp specific parsers).
I understand logging to much exceptions might pollute logging. I will 
have a look to this issue this week. Sincerely, Luc.

PR #27, opened 2015-11-30
added Russian synonyms

Stepanov-Sergey:


Orbiter:
Thank you Sergey!

Stepanov-Sergey:
source http://zhilinsky.ru/2007/11/26/synonims/ I just tried it. I can add many more words and I can add a dictionary of abbreviations, but to collect such a collection is hard work and time consuming

PR #37, opened 2016-01-04
Log files are commitable and shouldn't be

sudheesh001:


sudheesh001:
@Orbiter Could I have a quick review before I merge this, doesn't look like its affecting anything :smile:

Orbiter:
good point!

PR #38, opened 2016-01-06
Refactoring : use StandardCharsets instead of hardcoded charset names

luccioman:
Replaced hardcoded charset names strings with constants from java.nio.charset.StandardCharsets.
Advantages :
- easier charset references search within source code
- avoid Charset lookup each time a charset is used
- unnecessary try/catch against UnsupportedCharsetException can be safely removed

Non regression and compatibility checked on :
 - Linux Debian Jessie, openjdk 1.7.0_91 and oracle jdk 1.8.0_66
 - Microsoft Win7, oracle jdk 1.7.0_80 and oracle jdk 1.8.0_66


Orbiter:
Yes, thank you, good job! That was quite useful.

PR #40, opened 2016-01-14
Automatic crawling

Scarfmonster:
This is an attempt at creating an automated crawling mechanism. Using Solr grouping it retrieves a list of the oldest hosts in the index and then runs either shallow (depth 0-2) or deep (depth 1-5) crawls. Shallow crawls are an attempt to crawl as many domains as possible, while deep crawls try to retrieve as many documents as possible.

At deep crawl depth =< 3 this is an alternative to running very deep crawls in an attempt to crawl many pages. It should keep the crawling queue relatively small while requiring no monitoring from the user to keep crawling continuous.

It works best at bigger indexes which were created over the span of at least several days, as it will not try to re-crawl documents which were loaded in the last 24 hours (or more if set so).

### DIFF
--- a/defaults/yacy.init
+++ b/defaults/yacy.init
@@ -538,6 +538,15 @@ proxyURL.access=127.0.0.1,0:0:0:0:0:0:0:1
 proxyURL.rewriteURLs=domainlist
 proxyURL.useforresults=false
 
+# Autocrawl configuration
+autocrawl=false
+autocrawl.index.text=true
+autocrawl.index.meia=true
+autocrawl.deep.depth=3
+autocrawl.deep.recrawl=43200
+autocrawl.shallow.depth=1
+autocrawl.shallow.recrawl=1440
+
 # From the 'IndexCreate' menu point you can also define a crawling start point.
 # The crawling works the same way as the prefetch, but it is possible to
 # assign a different crawling depth.
--- a/defaults/yacy.init
+++ b/defaults/yacy.init
@@ -541,7 +541,7 @@ proxyURL.useforresults=false
 # Autocrawl configuration
 autocrawl=false
 autocrawl.index.text=true
-autocrawl.index.meia=true
+autocrawl.index.media=true
 autocrawl.ratio=50
 autocrawl.rows=100
 autocrawl.days=1
--- a/defaults/yacy.init
+++ b/defaults/yacy.init
@@ -542,10 +542,12 @@ proxyURL.useforresults=false
 autocrawl=false
 autocrawl.index.text=true
 autocrawl.index.meia=true
+autocrawl.ratio=50
+autocrawl.rows=100
+autocrawl.days=1
+autocrawl.query=*:*
 autocrawl.deep.depth=3
-autocrawl.deep.recrawl=43200
 autocrawl.shallow.depth=1
-autocrawl.shallow.recrawl=1440
 
 # From the 'IndexCreate' menu point you can also define a crawling start point.
 # The crawling works the same way as the prefetch, but it is possible to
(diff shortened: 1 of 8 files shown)


yacylover:
Great! Thank you very much. Is it also possible to determine the domains
with the fewest entries in the Index?  Please see

http://mantis.tokeek.de/view.php?id=75


Am 14.01.2016 um 03:31 schrieb Ryszard Goń:

Scarfmonster:
I don't think that is possible. There was an issue opened back in 2009, talking about the fact that faceting in Solr doesn't allow for reverse sorting, but it was never implemented into solr itself, and the patch is not working any more.

https://issues.apache.org/jira/browse/SOLR-1672

luccioman:
Great initiative! I wonder how far it could help growing the global DHT index and keeping it fresh...
It would also be interesting to distribute this kind of task across muliple peers to avoid unnecessary duplicate crawls.

Orbiter:
this looks promising! I have to tested it yet, let me see...

Orbiter:
ok, good work and this makes a lot of sense. I have some questions:
- why are the crawled urls not showing up in /Crawler_p.html ?
- After I started the autocrawler, it says that it needs a restart. But it works without. Why the restart?
- You added a busy task, what does it do, collecting new urls?
- is there a throttling, like: if the crawl queue is too full, the autocrawler busy task stops with feeding?

reger24:
longer term, I suggest to put this feature and RecrawlBusyThread (/IndexReIndexMonitor_p.html?setup=recrawljob) together, as it is a similar function just using crawldepth=0 to recrawl old documents,
what would only be a additional checkbox here (reload root).

Scarfmonster:
* **why are the crawled urls not showing up in /Crawler_p.html ?**
I create two default crawling profiles (robot_autocrawlDeep, robot_autocrawlShallow), which /Crawler_p.html does not display in the same way a regular crawling job/profile is displayed, if that's what you mean.

* **After I started the autocrawler, it says that it needs a restart. But it works without. Why the restart?**
The crawling profiles are created at the start of YaCy and not changed during it's operation, as I couldn't figure out a good way to edit the crawling profile after it has been created.

* **You added a busy task, what does it do, collecting new urls?**
The busy task is retrieving and adding the urls to the crawling queue. I reasoned that would be the most reasonable way to implement this, as it uses already existing parts of yacy to control some of the behaviour (how often it runs, load & memory requirements).

* **is there a throttling, like: if the crawl queue is too full, the autocrawler busy task stops with feeding?**
Currently it stops adding new urls if there are more than 200 in the queue. Perhaps this could also be editable.

Looking at /IndexReIndexMonitor_p.html it is indeed similar, the main difference with what I did is the autocrawl code always uses hosts as starting points, while the Recrawl reloads all old urls in the index. I am not really sure the behaviour of both is similar enough to merge them though, as someone may want to have both active. That is, to use Autocrawl to discover as many hosts as possible, and then use Recrawl  to keep index fresh.

yacylover:
It's a pity. Starting a crawl from a domain that has the fewest entries
in the index would be a great enrichment for the diversity of the
network. Since the index browser is displaying the amount of entries for
a domain it can be a possible source for that value.

Am 14.01.2016 um 12:25 schrieb Ryszard Goń:
> I don't think that is possible. There was an issue opened back in 2009,
> talking about the fact that faceting in Solr doesn't allow for reverse
> sorting, but it was never implemented into solr itself, and the patch is
> not working any more.
> 
> https://issues.apache.org/jira/browse/SOLR-1672
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/yacy/yacy_search_server/pull/40#issuecomment-171617543>.
> 
>
Am 14.01.2016 um 10:48 schrieb Stefan Meurer:
> Great! Thank you very much. Is it also possible to determine the domains
> with the fewest entries in the Index? Please see
>
> http://mantis.tokeek.de/view.php?id=75

PR #44, opened 2016-02-20
Created a translation file ja.lng

ImpactCrater:
I wrote a bit of translation to Japanese.


reger24:
Thanks for your effort and adding it as git pull request.

Issue #16, opened 2015-10-02
Trying to run YaCy from git on OpenBSD: "Error: Could not find or load main class net.yacy.yacy"


dertuxmalwieder:
... when running `./startYACY.sh -d` as root. :(

smokingwheels:
I get that with win  2000 1.82 it used to run. Its been awhile sine I have had it on.

bshouse:
Does the yacycore.jar file exist in the "lib" folder?

dertuxmalwieder:
Np.

bshouse:
I am not sure which package you downloaded, but I had to run 'ant' (from Apache Ant) in the directory where build.xml was located to have it create that file. Might be worth a shot.

dertuxmalwieder:
Hmm, I didn't have to do that on FreeBSD ... :(

bshouse:
When I downloaded the zip from github, I had to run ant. I downloaded the Linux package from the official website yesterday and just did: ./startYACY.sh 

In both cases it worked. 

I do not know how familiar you are with Java, but this error means that the file it needs in not in the classpath (either the command line or the environment variable). Do you have a yacycore.jar file anywhere in your install path? Maybe try to install the package from the website and see if it works any better.

Please let me know how it goes and I will dig some more.  Thank you.

dertuxmalwieder:
The package from the website works indeed. I wonder why. Is there an Apache Ant on OpenBSD anyway?

bshouse:
I am not sure if there is a package available, but I am confident it can be compiled from source. 

Thank you for fighting through and not giving up.

sudheesh001:
@dertuxmalwieder Were you able to compile from source ?

dertuxmalwieder:
Sorry, I was busy with things.

Meanwhile, I found `ant` (`pkg_add apache-ant`) to be working. I can't get a web connection over port 8090 though. :(

sudheesh001:
@dertuxmalwieder I am not sure if there's an `ant` by default on OpenBSD. If you're cloning from the github repo, you need to build the source using `ant` before running `./startYACY.sh`

I think this issue can be closed with an update to the documentation.

PR #45, opened 2016-02-24
Prevent start scripts to fail silently when yacycore.jar was not built.

luccioman:
Added a check for lib/yacycore.jar file, eventually followed by an error message and an exit status 1. 

Because one can easily forget yacycore.jar was not built, for example just after getting sources from github without and IDE or after a "ant clean" : it was then confusing to see  message such as "YaCy started as daemon process." and nothing happens.

### DIFF
--- a/startYACY.bat
+++ b/startYACY.bat
@@ -7,6 +7,11 @@ echo . >DATA\yacy.noconsole
 
 Rem Setting the classpath
 Set CLASSPATH=lib\yacycore.jar
+Rem Check core jar exists to avoid failing silently later
+if not exist lib/yacycore.jar (
+	Echo "Error : lib/yacycore.jar was not found! Please first build from sources using Apache Ant."
+	Exit /b 1
+)
 
 REM Please change the "javastart" settings in the web-interface "Basic Configuration" -> "Advanced" 
 set jmx=
--- a/startYACY.sh
+++ b/startYACY.sh
@@ -188,6 +188,13 @@ CLASSPATH=""
 for N in lib/*.jar; do CLASSPATH="$CLASSPATH$N:"; done
 CLASSPATH=".:htroot:$CLASSPATH"
 
+# check core jar exists to avoid failing silently later
+if [ ! -f lib/yacycore.jar ]
+then
+   echo "Error: lib/yacycore.jar was not found! Please first build from sources using Apache Ant."
+   exit 1
+fi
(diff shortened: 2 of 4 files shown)


reger24:
Hi,
in my oppinion, 
the startscript is not really used in build environment so not the ideal place to check for deletion during build.

reger24:
P.S. do you see maybe a way of general error checking of the start .... command.....

luccioman:
Hello, personally I used start scripts to test some dev I did : I used ant or maven to build a dist archive and to run it on target systems without need for an IDE. But at first something went wrong and on target computers start script said it had started YaCy but nothing happened...

I searched and tried for a more general error checking, but it is not easy as you can not rely on start command return code... Maybe checking for yacy.running file could be an idea?

luccioman:
Hi, finally still now I see no simple and reliable way to check status of a process started as daemon : I think waiting some time and checking status of some file or url is not really a good idea. After all we launched a process as background...

But when we launch the script with -d option, YaCy process is no more a daemon, and the error telling compiled class are missing is displayed. 
So don't you think a simple supplementary advice message could be useful. Such as : 
"** DEBUG      YaCy : execute startYACY.sh -d to start YaCy as a foreground process and get log messages in console. This may help you if you encounter errors when starting YaCy. **"

luccioman:
Hi, @Orbiter , @reger24 , what's your current opinion about adding an advice message about debugging or the initial idea of checking for lib/yacycore.jar? If you think it is not helpful I close this Pull Request without regret.

reger24:
Hi,
in my opinion, I'd not put much pure development environment things in a start script used for production. 
I'd expect a forgotten build is not a frequent issue to protect against and I'd keep it out.

On the other hand, if you feel  a consistency check (as many other systems do) of all important resources is the goal, I might change my above mind ;-)

cu

luccioman:
Ok @reger24 , I agree with you that this problem rarely occurs during the development process. 

But my idea was rather that it could be helpful for people who want to run YaCy straight from sources instead of a packaged version. Everyone may not be familiar with the YaCy build system or with development tools in general, so any complementary hint can help (see for example https://github.com/yacy/yacy_search_server/issues/16 or http://forum.yacy-websuche.de/viewtopic.php?f=8&t=5864).

Let me know if this changes your mind ;)

Best regards

PR #51, opened 2016-05-21
Fix multiple boost queries

JeremyRand:
Multiple Boost Queries (e.g. the default setting that YaCy uses for Boost Queries) are broken in YaCy right now.  See my writeup at https://gist.github.com/JeremyRand/39d0d77407d651539c7bb920b668d91e for details on what exactly is wrong.

This PR fixes the issue by allowing multiple Boost Queries to be specified, one on each line.

As a prerequisite, I had to refactor the escaping code that YaCy's config file read/write code uses, because it was failing to properly escape newlines (and its usage of RegEx was unnecessarily complex and difficult to read).  I haven't measured performance, but I think it's likely that the refactored code will be faster, because standard string search operations are much less computationally complex than RegEx operations.

If you need, I can provide the refactored escaping code in a separate PR.  I don't think this is needed, since it's already in a separate commit.

### DIFF
--- a/defaults/yacy.init
+++ b/defaults/yacy.init
@@ -1036,7 +1036,7 @@ search.ranking.rwi.profile =
 search.ranking.solr.collection.boostname.tmpa.0=Default Profile
 search.ranking.solr.collection.boostfields.tmpa.0=url_paths_sxt^3.0,synonyms_sxt^0.5,title^5.0,text_t^1.0,host_s^6.0,h1_txt^5.0,url_file_name_tokens_t^4.0,h2_txt^3.0,keywords^2.0,description_txt^1.5,author^1.0
 search.ranking.solr.collection.filterquery.tmpa.0=
-search.ranking.solr.collection.boostquery.tmpa.0=crawldepth_i:0^0.8 crawldepth_i:1^0.4
+search.ranking.solr.collection.boostquery.tmpa.0=crawldepth_i:0^0.8\ncrawldepth_i:1^0.4
 search.ranking.solr.collection.boostfunction.tmpb.0=
 search.ranking.solr.collection.boostname.tmpa.1=Date Profile: sort by date in descending order for a '/date' usage
 search.ranking.solr.collection.boostfields.tmpa.1=url_paths_sxt^0.1,title^0.1,text_t^0.1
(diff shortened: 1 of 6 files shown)


reger24:
Reading your test article, you are probably right,
that boost_queries (bq  param) need to be separated params,

After your hint.... reading the Solr wiki exactly.... it is in harmony with your test findings
https://cwiki.apache.org/confluence/display/solr/The+DisMax+Query+Parser#TheDisMaxQueryParser-Thebq(BoostQuery)Parameter

reger24:
I can confirm your finding & testing.
Only question for me is if a fix with introducing a \n newline or keeping the existing syntax and just use 
a split on " " space-char  { bf.split(" ") } is the best fix.

Right now, I'd prefer the later.

JeremyRand:
Hash: SHA256

I believe that it is legal for a single Solr query to contain a space,
which makes any such splitting problematic.  (For example, a query
might contain a multi-word phrase enclosed in quote marks.)  That's
why I opted for a newline -- it keeps the code simpler than the
alternative, which would involve checking whether the space is between
quote marks, and whatever other parsing rules Solr uses.

If you think I'm incorrect, let me know -- I haven't tested
space-containing Solr queries in any detail, so I might be wrong about
what Solr's syntax rules are.
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iQIcBAEBCAAGBQJXQgP8AAoJEAHN/EbZ1y06snwQAItytrbH6lkzOHwEHrhZDmNn
mOTXw35lopm93vq/sbHwliOedFbrWhQh8buTuOvaYqOHHJjZU9LDRKDDdXM069dy
ReDfHoGQAuBsuqR2ptwGzUgin1ckRxDDzTO8pcR4K6k0R0UN+5S2W0Bv0SUe/79X
B9NPtZYXfujsLQc2XSypzwM80Prp9cnWT6/PzxowlDtzxBM+F0eb+Qh4lHVUXhRP
HWROvIbMnNcgOT8WW/tZePF/7QNjAq2KjXLpcV0qEZLnkErg+Q+i4FjP/sgHbRKW
girFKdThaDHrOsh1ta3lOKk1WhwBGVPs1CGuPUIdf4D3NFiCwxv4171UKGYwaIBm
GrUVmZo43OcofIxci9NtLBIvtP1/ENAjY1q8rvgIXsqLFkAsQfwevb1YIbUNGvZE
byoYiRAQ5qEMT5XcNrjLbPmH8X+LwvNrBEoskOvXnDxyqStifB07CIjpSXt5FLz1
M/X5QQYXX9bSa7FFiZGQI2CXpt3C+7uleV3dxe+yWubgcP459+GD0oEsCwCTwCTV
bgUtXvS5G6qAfCOMCFsnazASR7rdVKXvEkzHjHHp/C18uezPpZ2XDp+GtVhIBjGV
Zz3pknAdGh4oiP315mwbEEz13lZ68TRYMSIxesPaA6fKfPQlQW35lHAUMCZaV5Vy
goEq6iAvS//j8Ytb2tTC
=VyhK
-----END PGP SIGNATURE-----

reger24:
You are right. a space split would restrict allowed syntax options.
Internally YaCy uses a bf param with spaces in picture search, which would be incorrectly splitted using space split.

So your fix using a line delimiter is the better choice.

Orbiter:
+1

PR #61, opened 2016-07-13
Deploy YaCy on Heroku

luccioman:
Made necessary adjustments to enable YaCy deployment on Heroku platform in one click.

All relevant information (various deployment methods, limitations, technical details) was added to a new file [Heroku.md](heroku.md).

All tests performed with a Heroku free plan account.

Details about current limitations were also exposed on [YaCy forum](http://forum.yacy-websuche.de/viewtopic.php?f=8&t=5810#p32972)


Orbiter:
great work! I tested this and it works perfectly.
There is one problem: how should I submit  base64-encoded password, that is a difficult thing. Can we add a hack so that a password submitted in the heroku deploy does the encoding itself?

luccioman:
Thank you @Orbiter for reviewing and merging!
I agree filling the admin password is not really straightforward... But I tried to document it as best as I could here : https://github.com/yacy/yacy_search_server/blob/master/Heroku.md#administrator-password.
Indeed the command described there could be run before launch, but I am not sure storing a clear text password in Heroku env variables is a good idea regarding security...
By the way I can look again for simplifying the necessary command to run, using for example standard Unix commands...

PR #66, opened 2016-08-12
Opensearch : fix for default ports and naming consistency

luccioman:
Handle correctly Opensearch description URLs when YaCy is bound on default HTTP and HTTPS ports, respectively 80 and 443. This extends improvements already made for mantis 669 (http://mantis.tokeek.de/view.php?id=669).

Also fixed Opensearch naming consistency between descriptor ('ShortName' field) and HTML search link ('title' attribute), causing incorrect behavior on Firefox as reported in mantis 679 (http://mantis.tokeek.de/view.php?id=679).

Test platform :
- YaCy peer running on Linux Debian Jessie
- Browsers : 
  - Firefox 45.3.0 on Debian Jessie
  - Firefox 47.0.1 on Microsoft Windows 10
  - Chrome 52.0.2743  on Microsoft Windows 10

### DIFF
--- a/htroot/opensearchdescription.java
+++ b/htroot/opensearchdescription.java
@@ -25,6 +25,7 @@
 // Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 import net.yacy.cora.protocol.Domains;
+import net.yacy.cora.protocol.HeaderFramework;
 import net.yacy.cora.protocol.RequestHeader;
 import net.yacy.search.Switchboard;
 import net.yacy.search.SwitchboardConstants;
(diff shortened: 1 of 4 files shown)


reger24:
Hi @luccioman ,
fyi: I consider and experiment for a while to have YaCy's custom HeaderFramework to implement HttpServletRequest (as all requests are originating from YaCyDefaultServlet). This would make alle HttpServletRequest methodes available in YaCy's servlets, incuding getScheme etc. and make the common ServletRequest standard available in YaCy Servets. This would also benefit this patch and make custom headers obsolete. 
For now, your patch is a good improvement.

P.S. hold up is currently the missing integration with HttpServletRequest and Container Authentication to also make the getUserPrincipal, login and authenticate work.

luccioman:
Ok thanks @reger24 , this is good to know and indeed this would certainly be useful in some YaCy's servlets.

PR #68, opened 2016-08-24
Fixed packaging for YaCy 1.90 on MacOS

luccioman:
Rely on startYACY.sh script as main entry point instead of JavaAppLauncher program.

Issues JavaAppLauncher had :
- jar and classpath resources are expected to be in YaCy.app/Contents/Java instead of YaCy.app/Contents/Resources/Java
- app data folder is retrieved in main yacy.java class with "user.dir" JVM System property. This is supposed to be the current working directory, but when running inside a MacOS app package, it is in fact the home directory  (same a "user.home" property), which does not contains the necessary defaults and other resources.

The first one could be easily fixed, but not the second one : so it is far more reliable to start YaCy using the startYACY.sh script.

Sucessfully tested packaging with ant task "distMacApp" and deployment on a MacBook Pro with Mac OS X "El Capitan".


Orbiter:
Wow, very good, thank you!

PR #77, opened 2016-10-04
Ranking Transparency.  

JeremyRand:
The exact calculations used to generate ranking can now be accessed via the JSON and RSS API's.  Currently only available for results whose Pre-Ranking was performed with Solr (not RWI).

### DIFF
--- a/htroot/RankingRWI_p.html
+++ b/htroot/RankingRWI_p.html
@@ -60,6 +60,12 @@
         </dl>
       </fieldset>
       <fieldset>
+        <legend>Explanation</legend>
+        
+        <label><input type="checkbox" name="localexplain" value="1" #(explain)#:: checked="checked"#(/explain)#> Generate explanation of how ranking was calculated for each result.  (This may decrease performance.)</label>
+        
+      </fieldset>
+      <fieldset>
         <input type="submit" name="EnterRanking" class="btn btn-primary" value="Set as Default Ranking" />
         <input type="submit" name="ResetRanking" class="btn btn-primary" value="Re-Set to Built-In Ranking" />
       </fieldset>
--- a/htroot/RankingRWI_p.java
+++ b/htroot/RankingRWI_p.java
@@ -103,6 +103,7 @@ public class RankingRWI_p {
     private static void putRanking(final serverObjects prop, final RankingProfile rankingProfile, final String prefix) {
     	putRanking(prop, rankingProfile.preToExternalMap(prefix), prefix, "Pre");
     	putRanking(prop, rankingProfile.postToExternalMap(prefix), prefix, "Post");
+    	prop.put("explain", rankingProfile.enable_explain ? "1" : "0");
     }
 
     private static void putRanking(final serverObjects prop, final Map<String, String> map, final String prefix, final String attrExtension) {
--- a/htroot/yacysearchitem.java
+++ b/htroot/yacysearchitem.java
@@ -197,6 +197,9 @@ public class yacysearchitem {
             prop.putHTML("content_faviconCode", URLLicense.aquireLicense(faviconURL)); // acquire license for favicon url loading
             prop.put("content_urlhash", urlhash);
             prop.put("content_ranking", Float.toString(result.score()));
+            prop.put("content_showExplain", theSearch.query.ranking.enable_explain ? 1 : 0);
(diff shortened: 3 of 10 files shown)


luccioman:
Hi @JeremyRand I tested your proposition and it appears to be working fine. 
But when enabled it adds quite a heavy load to the json or RSS result... Do you have some tooling to use the produced data, or is it just to perform manual check? I agree with the ranking transparency idea but to my mind it is quite difficult to read indeed. 
While potentially useful, shouldn't these data rather be added to YaCy logs with 'FINE' level?

JeremyRand:
Hi @luccioman ,

> But when enabled it adds quite a heavy load to the json or RSS result

Are you referring to the size of the generated JSON/RSS, or are you referring to an increased CPU/memory load?

> Do you have some tooling to use the produced data, or is it just to perform manual check?

Yes, my use case for this is a library I wrote called PyYaCySim, which accepts the following as input:

* The data generated by this PR for a given search result
* A set of Solr and YaCy ranking rules

And outputs what the ranking score would be for that search result with the specified ranking rules.  I also intend to add the ability to output what the derivative of the ranking score is, with respect to each ranking rule, which is useful for applying machine learning algorithms to optimize the YaCy ranking rules.

> I agree with the ranking transparency idea but to my mind it is quite difficult to read indeed.

It is somewhat difficult to read, yes.  The format that it outputs in is actually a pre-existing JSON and XML format defined by the Solr library, and is designed mostly to be machine-readable (e.g. by my PyYaCySim library described above).  That said, I've found that the JSON output isn't very hard to read by eye with a bit of practice.  (I agree that the XML output is difficult to read by eye, but that's mostly because XML-encoded data is inherently more difficult to read by eye than JSON-encoded data.)

> While potentially useful, shouldn't these data rather be added to YaCy logs with 'FINE' level?

My usage of PyYaCySim requires a JSON API where I can request YaCy searches via HTTP and get back the ranking data.  So logging it to the logfiles wouldn't really work for that use case.

If you like, I'd be happy to post my PyYaCySim code on GitHub (either before or after this PR is merged; up to you).  It'll probably take me a few days though.  Also the derivative calculation code isn't tested sufficiently yet, so all I can quickly post is the standard simulation code.  (Which I think is still pretty useful for people who are interested in tinkering with the ranking code.)

Cheers.

reger24:
w/o a additional tool I don't see who would consume the explain data (in this format). 
Except your planned tool, is there already any out there which can read the data/the format ?

luccioman:
Thanks @JeremyRand for your explanations. I was talking about the size of the generated JSON/RSS (I didn't noticed a big drop in performances, but didn't measured it precisely).
To my mind sharing your library would make sense if we were to merge this PR in YaCy. And as said @reger24 , maybe are you aware of any existing tooling that can already process this kind of data?

luccioman:
Hi @JeremyRand , I was just wondering : considering that currently you only add explain information to the results coming from Solr and not to the ones from distributed RWI, why not rely on the already existing YaCY "/solr/select" API? 
With output formats "wt=json" or "wt=xml", detailed explain information is already available when adding "debug=results" or "debug=all" to the query... You needed this explain information to be also available with the mixed peer-to-peer and solr search results?

JeremyRand:
Hi, sorry for the delayed reply.

I've posted an initial draft of pyYaCySim at https://github.com/JeremyRand/pyYaCySim ; hopefully it conveys a way that the data can be used.

> Except your planned tool, is there already any out there which can read the data/the format ?

The closest thing I can easily find on GitHub is https://github.com/solrpl/explain .  (It has a few forks which are newer than that repo.)  I think it probably wouldn't be very hard to make that tool and this PR play well together, but I haven't looked at that tool's code very closely, so I'm not certain.  At first glance it looks like that tool is using Solr's .txt format of the explain data rather than .json or .xml, which is a rather strange way to do it.  If that's correct, it's probably not too hard to modify this PR to support that format, I'd just have to look at Solr's source code and see what class is used for the .txt serializer.

> considering that currently you only add explain information to the results coming from Solr and not to the ones from distributed RWI, why not rely on the already existing YaCY "/solr/select" API? 

2 reasons come to mind:

1. YaCy queries remote Solr instances when doing a P2P search; this PR's output includes those results, whereas using `/solr/select` would only query the local Solr instance.
2. YaCy's RWI post-ranking changes the scores that are retrieved from Solr; this PR's output includes the post-ranking calculations.

PR #78, opened 2016-10-04
Show ranking in HTML UI.

JeremyRand:
Add preference (disabled by default) to show the ranking for each result on the HTML UI.

### DIFF
--- a/defaults/yacy.init
+++ b/defaults/yacy.init
@@ -849,6 +849,7 @@ search.result.show.hostbrowser = true
 search.result.show.vocabulary = false
 search.result.show.vocabulary.omit = 
 search.result.show.snapshots = false
+search.result.show.ranking = false
 
 
 # search navigators: comma-separated list of default values for search navigation.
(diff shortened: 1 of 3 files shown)


luccioman:
Hi @JeremyRand I also like the idea to give more transparency about ranking in YaCy UI. 
I just tested and it works fine, but to be truly useful I think this should be configurable from /ConfigSearchPage_p.html and not only through yacy.conf file or advanded config (/ConfigProperties_p.html).

@Orbiter would you agree on merging this PR? If so maybe Jeremy would like to add the related config option in /ConfigSearchPage_p.html?

JeremyRand:
Hi @luccioman , thanks for testing.  

> to be truly useful I think this should be configurable from /ConfigSearchPage_p.html and not only through yacy.conf file or advanded config (/ConfigProperties_p.html).

That's an excellent point; I agree that it would be beneficial to make it more easily configurable.  I'd be happy to make that change as well, either as part of this PR or as another PR (up to you).  Might take me a few days though.

luccioman:
Ok @JeremyRand , to my mind the most simple would be to perform the complementary changes on the same branch and therefore to update this same PR. And no problem, take your time. 
I just would like to add I personally like this proposal, but as I have been very recently included in YaCy developers team, I will let @Orbiter or other @yacy/yacy_developers do the merge if he/they want to...

reger24:
viewpoint:
The pure ranking number in result may be good for debugging but is imho otherwise relative meaningless for the searcher. But I understand the proposal has more in mind as showing A number.
Would be good to see the path for it, (added a starting point as discussion 
http://forum.yacy-websuche.de/viewtopic.php?f=8&t=5884 ).

P.S. I see the  /ConfigSearchPage_p.html  too as place for config (if we get more as e.g. score=385996 )

JeremyRand:
@reger24 I've replied to your forum thread, sorry for the delay.  Thanks for formalizing a lot of the relevant goals here.  I'll hold off on adding the option to the UI in this PR until a bit more discussion has happened in that forum thread.

PR #91, opened 2016-11-18
(WIP) Tradeoff between sorting accuracy and display latency

JeremyRand:
(Don't merge this PR, it's an experiment.)

At the moment, from what I can tell, YaCy waits a very short period of time, then drains exactly 1 result from the Solr and RWI queues into the output queue, and then checks whether the output queue has at least 1 result in it that hasn't yet been displayed to the user.  Naturally, this is true 100% of the time.  When YaCy sees this, it writes the best result in its output queue (i.e. either the single Solr result or the single RWI result) to the user.  It then repeats.  (See the section of code modified by this PR to see what code I'm talking about.)

The effect is that there is almost no correlation between the ordering of the results displayed to the user and the ranking calculated by YaCy; the dominant factor is which results got returned first by a peer.  Kind of useless as a sorting mechanism.

The easy hack to fix this is to add a significant delay right before the above algorithm runs, which gives the Solr and RWI queues enough time to receive a lot of results before they start being drained into the output queue.  The example code in this PR is a hacky way of doing this.

However, the problem is that it adds a long delay before the end user sees any results at all, and there is a tradeoff between the delay and the sorting quality.

My suggestions for a better way to implement a fix than this PR:

1. For API-based use cases that retrieve a single list, e.g. the XML and JSON output formats, provide a way for the user to choose how long to wait for results before the queues start being drained (probably an HTTP GET parameter is reasonable for this).  My guess is that 5 to 10 seconds is going to be enough time for an API user on a fast residential Internet connection (or at least, the first 100 results are likely to be properly sorted in such cases).  Users on Tor are likely to need more time.
2. For human-user use cases where Javascript is available, load each result via AJAX and use Javascript to insert it in the correct order.  Don't stop receiving results over AJAX when the result limit is hit; keep receiving them for approximately 1 minute.  Hide the least relevant results via Javascript once more results have been received over AJAX than the result limit.
3. For API-based use cases where the user is willing to call the API multiple times, re-sort the result list each time the API is called, so that the user gets more relevant results after waiting but they get some possibly-relevant results quickly.  Probably one way to do this with minimal code changes is to create a new queue, drain the entire output queue into the new queue, and then replace the output queue with the new queue.

If these suggestions are okay with you, I'm willing to try my hand at implementing them.  Or, if you have a better suggestion for how this should be handled, please feel free to let me know.

### DIFF
--- a/source/net/yacy/search/query/SearchEvent.java
+++ b/source/net/yacy/search/query/SearchEvent.java
@@ -1591,10 +1591,18 @@ public final class SearchEvent {
         }
         
         // now pull results as long as needed and as long as possible
-        if (this.remote && item < 10 && this.resultList.sizeAvailable() <= item) try {Thread.sleep(100);} catch (final InterruptedException e) {ConcurrentLog.logException(e);}
-        while ( this.resultList.sizeAvailable() <= item &&
+        //if (this.remote && item < 10 && this.resultList.sizeAvailable() <= item) try {Thread.sleep(100);} catch (final InterruptedException e) {ConcurrentLog.logException(e);}
+        if (this.remote && item == 0) try {Thread.sleep(2000);} catch (final InterruptedException e) {ConcurrentLog.logException(e);}
+        // this.rwiQueueSize() > 0                 : always wait
+        // this.nodeStack.sizeQueue() > 0          : always wait
+        // this.resultList.sizeAvailable() <= item : wait if (!this.feedingIsFinished() && System.currentTimeMillis() < finishTime)
+        while ( this.rwiQueueSize() > 0 || this.nodeStack.sizeQueue() > 0 || 
+                ( this.resultList.sizeAvailable() <= item && !this.feedingIsFinished() && System.currentTimeMillis() < finishTime ) ) {
+        /*
+        while ( this.resultList.sizeAvailable() <= item ||
                 (this.rwiQueueSize() > 0 || this.nodeStack.sizeQueue() > 0 ||
                 (!this.feedingIsFinished() && System.currentTimeMillis() < finishTime))) {
+        */
             if (!drainStacksToResult()) try {Thread.sleep(10);} catch (final InterruptedException e) {ConcurrentLog.logException(e);}
         }
         
--- a/source/net/yacy/search/query/SearchEvent.java
+++ b/source/net/yacy/search/query/SearchEvent.java
@@ -1592,7 +1592,7 @@ public final class SearchEvent {
         
         // now pull results as long as needed and as long as possible
         //if (this.remote && item < 10 && this.resultList.sizeAvailable() <= item) try {Thread.sleep(100);} catch (final InterruptedException e) {ConcurrentLog.logException(e);}
-        if (this.remote && item == 0) try {Thread.sleep(2000);} catch (final InterruptedException e) {ConcurrentLog.logException(e);}
+        if (this.remote && item == 0) try {Thread.sleep(5000);} catch (final InterruptedException e) {ConcurrentLog.logException(e);}
         // this.rwiQueueSize() > 0                 : always wait
         // this.nodeStack.sizeQueue() > 0          : always wait
         // this.resultList.sizeAvailable() <= item : wait if (!this.feedingIsFinished() && System.currentTimeMillis() < finishTime)


luccioman:
Hi @JeremyRand, to my mind making this waiting time value configurable trough parameter and/or configuration settings instead of using a hard coded value could be a good thing.

For the second point I am not sure to understand why the ordering taks would be better managed on the browser side... The complexity of mixing and ordering code would then be both in server and browser sides, potentially making it harder to maintain.

Another approach I was thinking of (but I didn't explore it that much) would be to simplify the ranking by only requesting the local peer Solr index AFTER adding RWI results to it, relying on the ["Near Real Time Searching"](https://cwiki.apache.org/confluence/display/solr/Near+Real+Time+Searching) capabilities of Solr to get results fast enough. Maybe it is not really feasible... did you had a look in that direction?

JeremyRand:
Hash: SHA512

luccioman:
> Hi @JeremyRand, to my mind making this waiting time value 
> configurable trough parameter and/or configuration settings
> instead of using a hard coded value could be a good thing.

Okay.

> For the second point I am not sure to understand why the ordering 
> taks would be better managed on the browser side... The complexity 
> of mixing and ordering code would then be both in server and 
> browser sides, potentially making it harder to maintain.

The problem with doing everything on the server side is that the
server is outputting a single stream of results, and that means that
once the server outputs a result to the stream, it cannot later insert
another result before the result that was already output.  So if a new
result is received remotely that turns out to be more relevant than a
result that was already output to the browser, the server cannot fix
this.  In contrast, using Javascript to re-order the results as
they're received from the server would allow a result that was
received later to be inserted before a result that had already been
received.

The alternative is, of course, to wait a long time for results to come
into YaCy before YaCy send any of them to the browser, but that means
that the user has to wait a long time before seeing any results at
all, and there is a tradeoff between the delay to see any results and
the accuracy of the sorting that can be done purely on the server
side.  Using Javascript to re-sort the results allows the user to see
*some* results in the browser instantly, and if they wait a short
while (say, 2 seconds) they'll see more relevant results, and if they
wait a little while longer (say, 5 seconds), they'll see even better
results.

The Javascript code shouldn't be complex.  I notice that YaCy already
includes jQuery; it's quite simple in jQuery to prepend an HTML
element to an existing HTML element, so basically the Javascript code
would iterate through the results that have already been displayed,
and when it finds the first displayed result that has a lower ranking
than the incoming result, it prepends the incoming result to that
displayed result (and removes the lowest ranking result that's been
displayed if the requested number of results have already been displayed
).

> Another approach I was thinking of (but I didn't explore it that 
> much) would be to simplify the ranking by only requesting the
> local peer Solr index AFTER adding RWI results to it, relying on
> the ["Near Real Time 
> Searching"](https://cwiki.apache.org/confluence/display/solr/ 
> Near+Real+Time+Searching) capabilities of Solr to get results fast
> enough. Maybe it is not really feasible... did you had a look in
> that direction?

This seems like an orthogonal solution; the problem that I'm
describing is that there's very variable network latency for when
results are received, and that waiting for all the results to be
received before we try to sort them will result in a long delay to see
any results (just because of network delay, not due to anything Solr
is doing).  I hooked up YaCy to an intercepting proxy (for an
unrelated experiment that deserves its own GitHub issue), and I was
seeing some Solr Select responses coming from remote peers quite a
long time after the initial search was started.  I didn't time it
explicitly, but something like 10 seconds delay to receive a response
from a peer was encountered on a regular basis.  This was on a
residential cable modem connection; Tor users will have even worse delay
.

That's not to say that Solr isn't also causing some delay -- just that
the network delay itself imposes constraints on the system that makes
it unfeasible to wait for all results to be received before we display
anything to the user.
-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJYM1KgAAoJELPy0WV4bWVwthoP/j8Lqb+bjy3X+14GAu6sDPyC
BGyLbPVtqKQDXBa+OSqvb6XodpCO9WtbnIP1EAeAVQYJLr1VyAW4swMRFxHn3t9/
yRbdZWVTayfmLNL9jNc0KoMZwXH47XKhdyhVXhdZSuQM0n3P+X11JgLufK5dpUpm
MyeCFjAhp76hJvwduF2df5GCMWV2fLZIXOGNWkYPb1xlb4IgTHdjRQD1QRS9Y/1i
wZ2hkvB0OMEOPO4G0VtCC4j3jg7e516Jw1RsG+VWtnTd9dEtFOoFqHqoLFYK6uuW
v40KETU7UolakfpinPuEcEcRtfh5Dv+EdyDRZYKPK77r1uoGCryIhnQkQFN+52Kh
gD37uufTw1eSFammRJArlPIYt6LFcVDs0ceJbsPmPU6D5FatCklxuWfC+8EJ6jmg
Xwu6x967LfkOxTp1cHfeyWLA4bc2dSV9H9dBBAbUKeyarLM1ZdDvTeQ7q77Nbhrf
4PxnL93zq2sQDtpcM4oWu1/RuELaWWHSr5ufSimKFOsBxhHHxewty4Sv9KQESEXP
+vmWoF9RvewyXJT239r8SkYAsupQHc4ihL+8pvV6fDR7MQ4VK4t4eCJPldzRCNeN
j8xUF+YMF+ZupoUIvErzOa13sieUE1yxrIIZeo/8nXypKo2rgUAnyF0amlD5Tqv/
hJjGpmFSerjc7xssSNLk
=W/Fj
-----END PGP SIGNATURE-----

luccioman:
This sounds like reasonnable arguments to me. 
I had also thought about results post-fixing in JavaScript as a possibly useful step in image search, for example to remove remaining HTTP 404 error status results or wrong MIME types.
But I guess we have to be careful not to introduce unwanted flickering in the results list when many ranking resorting occur...

JeremyRand:
Hash: SHA512

luccioman:
> This sounds like reasonnable arguments to me. I had also thought
> about results post-fixing in JavaScript as a possibly useful step
> in image search, for example to remove remaining HTTP 404 error
> status results or wrong MIME types. But I guess we have to be
> careful not to introduce unwanted flickering in the results list
> when many ranking resorting occur...

Okay.  Would you like me to try to produce a prototype of the
Javascript-based re-sorting so that we have something concrete to
evaluate?
-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJYM+zBAAoJELPy0WV4bWVwDvUP+gOOKhJWlmz2tm7klgRqpsQB
M/yxlcMNyufbS2K7r0Y/dIZcxbytNDet7T10PtC02h9KnmIsU4fQSGsdWbnY/cWr
APkyeT4QTSdHZXUj1TbdspEG5yqKNFiynuzpK9E0NGfcUwaZ+qakhemyTt0Gwl1g
3tB9G9uXVowd7eZw0r6djuwv32kEwGNp3WMbRum8HEXDEMEoY1QZKzR/kLFmGIuB
XfxZ6bno7S1pa95uMUZro9mLWCeACXGKqybczkNqu128F8xb8W+CZYeuAJTtS+GY
9CERXNnvN9ENUZZNwemZp/s37BCaVVUb4NtD855NryKwPwARORB+VbcJ+IDUcWQ7
AxSx1U1tkeRxIluza516J8dUjS9AlAMTLZbIdgXMOVKCX/ZC65/A/a+hH2ZbnoT/
Mv/YkASpk93r3snt1/ARAuWPoaSR5GvhbFomt2a2P09e3+rPEzfRKbXkGP27cIhA
f2hJ4qidqHzWNZQBGfdzqZTTaRScLQ8TqqA8qNBgOhzZFkv9p1qJW3J1nDOl6Qau
AG438gRtMNlTRNxEqXpb2BvRZhTKJH17XHxkqg2HgAGgLpvnjymupYrkKiljCoQ/
at2nyh8ULIwGJgHxVIZfMd1vhx8z9pJAhARJGie16wVizwg8sWEaY3wAoF4Oa7gs
k3IC2eb/RCZHDMiJXmh9
=hGaY
-----END PGP SIGNATURE-----

luccioman:
Hi @JeremyRand , if you want to have a try on this point of course I would be glad to test your prototype.

smokingwheels:
I like it the way it is. Even on my P4 the results come. There very good if I am sorting by date I know to leave it alone for about 10 seconds.
I do have a long block list that seems to remove some spamming sites hence the results are better.
My Peer is 4040 and may take some time to load if you want to try it. 
@JeremyRand 
@luccioman

luccioman:
To my mind the relevancy of results is in the end quite subjective and difficult to evaluate, even more in the context of a peer-to-peer search... I guess each YaCy peer owner has his/her own custom strategy to get the desired results. 
But personally I am still curious to see what Jeremy work could bring, and whether it can be helpful. Of course such a feature should be optional. 
Best regards

smokingwheels:
@luccioman Ok thanks for the reply.
Does @JeremyRand have a live server we could test and discuss?

JeremyRand:
Hash: SHA512

Hi,

I've been tinkering around with Javascript-based re-sorting.  I have
an initial WIP prototype if anyone would like to try it.

https://github.com/JeremyRand/yacy_search_server/tree/javascript-resort

It is definitely not ready for merging yet.  On my to-do list before
considering it a candidate for merging are:

1. Make the behavior optional, since my guess is that screen readers
won't handle it well (and browsers that don't support Javascript
definitely won't handle it properly).
2. Make sure that all the code paths work as expected (I haven't
tested edge cases at all).
3. Figure out how to handle multiple pages of results (I'm pretty sure
that that functionality is totally broken in my branch).
4. Fix the result statistics at the top of the page (in this branch
they're definitely wrong).
5. A lot of results are appearing and then immediately disappearing at
the bottom of the list; this is almost certainly a bug in my code.
The intended behavior is that results less relevant than the
Nth-ranked result (where N = the number of requested results) should
not be shown at all.
6. Rebase/squash and other Git cleanup.

However, notwithstanding the above issues, I'm curious whether people
like it.  In my initial testing, the subjective ranking quality is
dramatically improved with this change, and the most relevant results
sometimes don't arrive from P2P for several seconds after initial
results display.  Similarly, I can empirically verify that the results
are now accurately sorted by the YaCy ranking function.

I wasn't sure exactly how to best do the animation; it's possible that
a different animation style or speed would be better.  FWIW I think
this animation style definitely does the job acceptably well.

@smokingwheels While the definition of an ideal ranking function is
definitely subjective (and it's great that you like the results you're
getting), it is an objective, empirically observable fact that the
results being shown by YaCy do not accurately reflect YaCy's ranking
function, and are mostly instead a function of network latency.  It is
not realistically possible to improve YaCy's ranking long-term unless
that is fixed.  Of course, the choice of how to fix it is an
engineering decision where there may be disagreement -- I personally
think the three suggestions I made previously are useful (depending on
the use case), but maybe there are other fixes that I haven't thought
of... maybe you have a suggestion there?

Cheers!
-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJYP4oYAAoJELPy0WV4bWVwHUMQAMeOn1nQTPxGqyAfHhJz26Eq
r6WXH5xW7wMano7ocb4453uVsO4sEK4f6ynld2mwTspWE6GqEEmew0pxyHsvybEc
mF+iCUQtEk/CU5b8nmj73WAlMHTygmUg7/YMxAGp7FvxTKLCVro/PnjedQ3yOH28
+n/XiYDEEHoK9NysGiRpN3TcpDCKeRP0JCgAbaqoHilTDrKNflyvFZ7ePN4YthaK
yKWN/dS2EwhpwYZDSSuyNBVMTYgb/EwjhIULWVXGKofMWoxzFGmBsmsnv2Ep6qv8
b8zh7kq8XAYCVhueV80Dpq1LfN+sC+xKaYYLOEfbLhxlhNQBFnut0a8R0NKsn1jh
nkAFDjwjvSqzVy9mgQjlVqZdgqwBT9ylQx9u0AztCXXyQlSMB5iCpzg9dK+D20p5
bIi9NVMgo/xmu14gpPtq0Tdg/hHwarjRWBEFONZwtka2EYAX7ZqSMWm+vST/rM7Z
J+v7PcQEAEMlLXLkcT7hTaTfqm3qJ1J3qj8o2tvs69vbWQ9a7yzrnkKyDny6swFt
oK8+YalhiFkVxyO8HGgBEc2UzvKW20pfOHa+7Aak5vz2iq/bpqi1KUzUzSXBy1Cu
hx6f5Yntg72AvNH0J0IYO6sZEQOXtlnaCtxVDUbKWJP6H1d3JbpQTjb9duOc3f65
pd3+5XD+0djP4SJa6BSl
=HDTf
-----END PGP SIGNATURE-----

oyvinds:
I personally would not mind if getting YaCy search results took 10 or even 20 seconds to be presented

IF they actually came out ordered by something that could accurately be described as RANDOM.

A message saying [Gathering search results] with a progress bar or count-down or just [Gathering search results, this takes 20 seconds] would be fine. The actual value could be set on the /ConfigPortal.html page.

I do see the argument for client-side sorting but I do not like it. My personal preference for search result pages is https://duckduckgo.com/lite/?q=microsoft+tay+kill+humans - that is, I like a minimal page with no JavaScript and as little as possible beyond the actual search results. I do not work in a museum where I need to demonstrate the Internet on a Pentium4 during visitor tours but I do sometimes search on both cheap chink cellphones and pidgin-like Internet connections. 

Something related to this to think about: YaCy _could_ have a /lite/ version for /yacysearch.html which was totally plain. This way /yacysearch.html could have all sorts of client-side JavaScript for desktop users with not-museum-standard computers and EU-style Internet-connections (100mbit not 2mbit ADSL like they have in third world countries like the US) and a /lite/yacysearch.html with no JavaScript and no left-side sidebar and no nothing but search results and a search input field.

JeremyRand:
Hash: SHA512

@oyvinds If I'm reading your message correctly, it sounds like you're
not opposed to Javascript-based sorting being part of YaCy, you just
want it to be optional (and you would keep it disabled for your
usage).  Am I understanding that correctly?

If so, I fully agree that it should be optional.  Some users don't
want to run Javascript for whatever reason, some users use screen
readers that won't handle it well, some users are accessing YaCy
remotely from slow machines, some users are accessing YaCy via an API
where Javascript doesn't make sense.

The only reason it's not optional in the branch I linked to is that
that branch is an experiment, to get feedback on whether the
implementation works; it's missing a lot of needed stuff before I
would recommend it for merging.

FWIW, I personally would not want to wait 10-20 seconds to see any
results, so Javascript-based sorting fits my workflow very well.  It's
very subjective.  We both get what we want if it's optional.

Cheers.
-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJYR0XMAAoJELPy0WV4bWVw054P/RC6McoCjgAfWxuIN2tVNmUl
K0HhmKPW76nn6y132SaDbQoG6tAN9r8fr7Tuoe0ZRRStaqelo8+1LfNw07Bn5zwL
HQdCo8UGyF2yzv+yVXjyCoy6DOe32qfTZCM6xPrko8jOBRi+mnDyO8puokQPunUd
EUKfVHTpTN7KdQOciPTokOJNIZm1B62Y3hH1qrWg3mgotrZf+jk01FnS8f2axh28
eU7ULXlnIRG4egGNKUAsFUeIB5TEb1annZ7cAA08H/KldesMB7UYt6A3AGvOV7bj
cw46oy6OwMIQBP/pxEqpJPH9kVpLW27LZKytry1KBDm7eg4XA5hxeSLDKC7MU07+
EtZonCZqIxQ+G+V+MgXPR3uJyvv3EAoxcYyWpWlLP4lMFh2KIMcuOKpeAfS9jxzq
X/VdpG3Zm3msY0qktkEB54jjyWt1BfklnmflqHN3qQtOZiJu1Yhc0ecFV1MKqeYx
qf4mYozib1CEkXdIHAfgNaqZ2CVyfh3VPratvKw9ovnQslgqN+W6wtQDAhmQm+ob
t1yJ/L8uZ17ydoJdt3xS2r958uUtM+A/xFHVMoItACzE50nJM8Twmwa1XCtAU62n
maFpskAYQudGi7mCVml3Vcjta8b5MEI3MKtYpEcd9cdds4m9HrqEtAFgOD5qYjM/
YDXJGUijc8ewbfrU7BzM
=JY5T
-----END PGP SIGNATURE-----

oyvinds:
DuckDuckGo's normal search result page is https://duckduckgo.com/?q=javascript&ia=web with JavaScript everything like that. It works fine on desktops.

For mobile devices with a weak CPU or a small screen and old computers there is  https://duckduckgo.com/lite/?q=mobilefriendly which has no JavaScript and basically just the search results.

I think it would be easier to get everyone on board with heavy client-side JavaScript if YaCy had a similar solution with two different pages. Like http://yoona.everdot.org/yacysearch.html?query=test for desktop search and http://yoona.everdot.org/litesearch.html?query=test or something. If not /litesearch.html then perhaps /lite/yacysearch.html but you get the idea.

As for waiting 20 seconds before getting search results, such a solution would have to be configurable. Some would want 4 seconds and others 20. I am just saying that I would rather wait and have the results sorted than have results presented in a seemingly random order. YaCy is so hard to install and setup anyway that everyone who goes through all that would be able to configure such a setting.

 Jeremy, thank you for your hard work on this.

gardner:
@JeremyRand are you still working on this? Is there any way that I can help out? It sounds to me like a useful change. I could see it having three modes:

1. Disabled
2. Update UI in realtime
3. Update UI after a certain condition

Personally, I really can't handle things that I am trying to read move around on a page so I would prefer number three with a specified timeout or maybe a hotkey for the condition.

Alternatively I could see attempting to communicate rank using some design mechanism other than order on the page. I am not a designer but I could imagine this being implemented using font-weight, or background color.

I just cloned the repo and I'm building your PR.

gardner:
I just noticed #104 I'll move the conversation over there.

JeremyRand:
As I understand it, solutions 2 and 3 have been merged to YaCy.  Should we close this issue?

luccioman:
@JeremyRand , yes both points 2 and 3 have been merged. But point 1 (configurable initial wait time for API-based or without JavaScript P2P search) could also make sense and may also be considered. 
Are there some people here (still) interested by this feature?

oyvinds:
Just a tip for anyone setting up their own little thing, adding JeremyRand's 61084f8522c259295e6363239987bfe3ea6a1ce3 helps a whole lot when it comes to getting more useful results. If you go "System Administration" and "Debug/Analysis Settings" and turn on "Show search results scores" you'll see that YaCy currently presents results ordered by the first responding peers results sorted by score and then the next peers so higher ranked results from slower responding peers get pushed down. This makes the result list almost random both on the global yacy network and also when testing on a private LAN with a few peers. Adding a delay of one second makes the results be listed in the order of their rank not randomness on a LAN. Ideally YaCy (or a fork of it) should do a another sorting step which sorts by ranking using all results that are in by that time right before the results are presented.

PR #93, opened 2016-12-01
Allow disabling Javabin encoding for remote Solr queries.

JeremyRand:
YaCy configures Solr to return results in Javabin encoding rather than XML.  This is good from an efficiency standpoint, because Javabin is much more compact than XML.  However, Javabin only has an implementation in Java, whereas almost anything can process XML.  This is problematic, since as a Free Software program, YaCy should try to interoperate with other software as much as possible rather than locking users into Javabin.  In practice, this caused me a significant amount of frustration when I was trying to do experiments with the debugging tool mitmproxy, which (as a Python program) can't parse or serialize Javabin.

This PR adds an optional setting for enabling Javabin.  It defaults to true (meaning that the current behavior is preserved), but if it is explicitly set to false, YaCy will ask Solr to return results in XML (so that the protocol is interoperable with software that doesn't speak Javabin).

I wasn't sure where the best place would be to expose this option in the settings UI, so right now it's only accessible by changing the raw setting.  That's not really a good thing, since I doubt many users will notice the raw setting.  If anyone can offer suggestions where it would best fit, I will make that change before it's merged.

Cheers.

### DIFF
--- a/defaults/yacy.init
+++ b/defaults/yacy.init
@@ -1217,6 +1217,7 @@ federated.service.solr.indexing.sharding = MODULO_HOST_MD5
 federated.service.solr.indexing.lazy = true
 federated.service.solr.indexing.timeout = 60000
 federated.service.solr.indexing.writeEnabled = true
+federated.service.solr.indexing.binaryResponseEnabled = true
 
 # temporary definition of backend services to use.
 # After the migration a rwi+solr combination is used, the solr contains the content of the previously used metadata-db.
--- a/source/net/yacy/cora/federate/solr/connector/RemoteSolrConnector.java
+++ b/source/net/yacy/cora/federate/solr/connector/RemoteSolrConnector.java
@@ -27,6 +27,8 @@ import java.util.ArrayList;
 import net.yacy.cora.federate.solr.instance.SolrInstance;
 import net.yacy.cora.federate.solr.instance.RemoteInstance;
 import net.yacy.cora.federate.solr.instance.ShardInstance;
+import net.yacy.search.Switchboard;
+import net.yacy.search.SwitchboardConstants;
 
 import org.apache.solr.client.solrj.ResponseParser;
 import org.apache.solr.client.solrj.SolrClient;
@@ -52,7 +54,7 @@ public class RemoteSolrConnector extends SolrServerConnector implements SolrConn
     public RemoteSolrConnector(final SolrInstance instance, final boolean useBinaryResponseWriter) throws IOException {
         super();
         this.instance = instance;
-        this.useBinaryResponseWriter = useBinaryResponseWriter;
+        this.useBinaryResponseWriter = useBinaryResponseWriter && Switchboard.getSwitchboard().getConfigBool(SwitchboardConstants.FEDERATED_SERVICE_SOLR_INDEXING_BINARYRESPONSEENABLED, true);
         this.corename = this.instance.getDefaultCoreName();
         SolrClient s = instance.getServer(this.corename);
         super.init(s);
@@ -61,7 +63,7 @@ public class RemoteSolrConnector extends SolrServerConnector implements SolrConn
     public RemoteSolrConnector(final SolrInstance instance, final boolean useBinaryResponseWriter, String corename) {
         super();
         this.instance = instance;
-        this.useBinaryResponseWriter = useBinaryResponseWriter;
+        this.useBinaryResponseWriter = useBinaryResponseWriter && Switchboard.getSwitchboard().getConfigBool(SwitchboardConstants.FEDERATED_SERVICE_SOLR_INDEXING_BINARYRESPONSEENABLED, true);
         this.corename = corename == null ? this.instance.getDefaultCoreName() : corename;
         SolrClient s = instance.getServer(this.corename);
         super.init(s);
--- a/source/net/yacy/search/SwitchboardConstants.java
+++ b/source/net/yacy/search/SwitchboardConstants.java
@@ -302,6 +302,7 @@ public final class SwitchboardConstants {
     public static final String FEDERATED_SERVICE_SOLR_INDEXING_LAZY         = "federated.service.solr.indexing.lazy";
     public static final String FEDERATED_SERVICE_SOLR_INDEXING_TIMEOUT      = "federated.service.solr.indexing.timeout";
     public static final String FEDERATED_SERVICE_SOLR_INDEXING_WRITEENABLED = "federated.service.solr.indexing.writeEnabled";
+    public static final String FEDERATED_SERVICE_SOLR_INDEXING_BINARYRESPONSEENABLED = "federated.service.solr.indexing.binaryResponseEnabled";
 
     public static final String CORE_SERVICE_FULLTEXT            = "core.service.fulltext";
     public static final String CORE_SERVICE_RWI                 = "core.service.rwi.tmp";


luccioman:
Hi @JeremyRand I would have two comments :
- to my mind the current RemoteSolrConnector constructors already have a boolean parameter controlling the encoding of the exchanges with the requested index (which can also be the local embedded one). There is currently no dependency between the RemoteSolrConnector and the Switchboard and this is a good thing. If we want to modify this parameter, to my mind it should be done in the calling code as done [here](https://github.com/yacy/yacy_search_server/blob/master/source/net/yacy/peers/Protocol.java#L1077) for example. I agree it will imply more modifications, but will stay more flexible and generic, if for example we would like to configure differently the encoding between our own index and the encoding used in the federated search.
- maybe a good place to add this configuration setting in the admin UI could be in "Index Administration" > "Index Sources & Targets" (/IndexFederated_p.html)

reger24:
I don't really understand the benefit to change the internal response communication and how that would improve  
> interoperate with other software

For interoperation, rss, opensearch and still even /solr/select?q=xyz  are as xml style responses avail. to consume xml
and federated search connectors to process xml style input.
I'm a xml fan too, but here .... with the only internally processed response, stay efficient is fine for me.

luccioman:
As far as I understand, the benefit is rather to be able to debug/analyze remote Solr queries occurring during the federated search process with standard network analysis tools rather than with a Java IDE/Debugger. 
To my mind being able to switch at any time to a human readable exchange format is a plus, and  I guess it can be more convenient for anyone not familiar with Java tooling.

JeremyRand:
Hash: SHA512

luccioman:
> As far as I understand, the benefit is rather to be able to 
> debug/analyze remote Solr queries occurring during the federated 
> search process with standard network analysis tools rather than 
> with a Java IDE/Debugger. To my mind being able to switch at any 
> time to a human readable exchange format is a plus, and  I guess
> it can be more convenient for anyone not familiar with Java
> tooling.

Yes, that's correct.  In my case, I was trying to debug what would
happen when I triggered edge cases in the remote Solr query protocol;
I was using the tool mitmproxy to do this.

(I will address luccioman's technical points from Dec 2 as soon as I
can.)
-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJYRddyAAoJELPy0WV4bWVwvpEQANqyNgGBtLxxo11f8IRK1ouT
qsqc9vQI3lflND3lhPKH/KrcYSFp+SUVUXf1mPOfj/plGvqxNdcYJk34M096dVnq
Nx36WqAdvxPvWcpv2E3zVlLiC2kvXdVLPt+dm0PNr0G71RbAmSgznoMRQD06nsOC
k1+BkbNzCJxUByqqiD6CilppGwM/IDvlQzY4kaF5QLjEJvSrK0AF/ttP07pjuAXU
5c0Pm+ohdI78n4OesOV1r5y40CSSQNhnzQCis7jYktB0xKwvDGQ5JDBajvlYygIQ
c9cmCRysFZcCLJpYSmLfz4ixU0mKpe+NmID8b9X1FNtr9WFd1MSAF7bIXYbIwXaM
6LHYHSPeOL+bV9wuuzM7oKn4jnnklmz7+tR4uCCBh820fqUCy3pb06freUAAUFdl
VCHmNNh8hZ1tyXMuqSSGLqNMTIFsE2qNKXzTL45DZydCZK4Gftj2QY1l74r7gVpB
X7rpgc6Hd4Bgt1h1Ga7q0Q1WHy6K0jxrfrkhbaoOq9a6swNTSLq2lDbo8y+3KgMp
jO8H2jc0fWy0BdRpxaIO9gi/z+Sf2wy7JkxmB+r8ws+vP1ECcgPeFKaW3XopaMUW
0PbwmYrb9KPRArfEKN3ahiVffC+jNPk9IeEdZQk9lL1X8NGuwzKjGI1RPF0VpTce
JRSDQjDe2Yr/rOfi/qH6
=fFll
-----END PGP SIGNATURE-----

reger24:
> I wasn't sure where the best place would be to expose this option in the settings UI,
Then (with the aim to debug or interoperate with tools to debug), to not bother users (not interested in looking under the hood) with knops and switches the one best place IMHO: is under
System Admin -> Advanced Settings,  maybe even create a page "Debug & Log Settings" to group this and possibly the other existing (and coming) debug switches up at that place.

luccioman:
Agree with @reger24 , a new group of settings related to debug/log/analytics would be a good idea

JeremyRand:
Hash: SHA512

luccioman:
> Agree with @reger24 , a new group of settings related to
> debug/log/analytics would be a good idea

Yes, I agree.  Average users will avoid such a settings page so it
won't bother them, but users interested in debugging will be able to
find those settings much easier if they're in a dedicated page than if
they're buried in the raw advanced settings page.

What other settings, if any, would be a good fit for this page?

-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJYRnrSAAoJELPy0WV4bWVwGr4QALWeRaFCjZFOyNXXB/QpEYGw
8tMhCDMAK/Lz0n31wtoSaFbDd0zA3P3EptNel28r3FLosd56iT7xUBf32deKrXtc
fPkXysF8C51/bUCKpzoyOD0EsTxaL4zdZC+Vu6xL/fPeakc+OyhOPhBCdc6zu/AN
s+Y/SaLeN9YqoQzo6ClLiR5wRZJPqw8IVVM4VDYC5GritpTS5or9F2aMTDvSKbTG
/vK2WUWs5ffWgUPhv9YlP6/LCqkHi10HW14o5EH4dxdO082DfLdKNF/vcu/aZwBj
BbdeMcui+TQGOT8hbvdGGgO1CbzfxpP/+7fQKML4FVQCEF5/R4h3E1o/uqoohQM6
OosHNHVeud6dlnE0Puq/1/GMzs4oxl7vmoMupsk0PsuaJpq+tOF4iLE6gBkCQkU/
HunFikpd6AaohLKc7O9Cd1m7ZPbnrWSaJAEHpo4BD2NCEhz568ssjy0wkMNUP32k
lgpma6Ta1nM6PFw1qATuY87vz9QXLwqCrE7Jo6GBLjZCxNkzATFJ/l451ve2n6be
LX8sFmFheq307mQRV1y/waOZ/frZJZqXDsSOsZvtdb/KVwfCawkDSQy4eorOgcw2
zGubM8b+hVkqI/vWb8chayOLyVgWGY9g0EIZ2epaY2hGWywb0TcddWyamacDnTeF
nh2F00lkZaSy5Rq+zsZy
=vOMP
-----END PGP SIGNATURE-----

reger24:
A start point could be the none UI switches
https://github.com/yacy/yacy_search_server/blob/master/source/net/yacy/search/SwitchboardConstants.java#L329
as well as  things around the prev.  ranking logging topic.

JeremyRand:
Hash: SHA512

luccioman:
> to my mind the current RemoteSolrConnector constructors already 
> have a boolean parameter controlling the encoding of the exchanges 
> with the requested index (which can also be the local embedded 
> one). There is currently no dependency between the 
> RemoteSolrConnector and the Switchboard and this is a good thing. 
> If we want to modify this parameter, to my mind it should be done 
> in the calling code as done 
> [here](https://github.com/yacy/yacy_search_server/blob/master/ 
> source/net/yacy/peers/Protocol.java#L1077) for example. I agree it
> will imply more modifications, but will stay more flexible and
> generic, if for example we would like to configure differently the
> encoding between our own index and the encoding used in the
> federated search.

That's a good point.  I think I agree; I will look into making the
requested changes there.  Might take me a little while, things are
busy here this week.
-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJYUQj0AAoJELPy0WV4bWVwX60P/j/O6rvH5VcUydi66uXN1hm0
ndk1KHpXWU/9A4RlMYPpPy9RB4CTR0fWjy/fpQKyekZ+Dw1YwHwZ4TcTFAEx5HIB
Gs395hbNevhT1MF1XpQRwP72uisgNBcp4pDJKrJUNTSVYSbWe3PgPUm5B38DSfLy
zOWsYxKJ0tqfU/WhqcJTSC6ziuIHGxoybJZVX3Jgyn2OWCf4Rmqw23z3igR9rhvH
BuSXpflyLcFG5eMd4SH38he0J7TlIoQNlstMfGTfzq6za9TvImlneKGG/F0WMgGX
lGZoOaMJ1lr/Ys59WRHbRdrz1X0uaXtxN/svOo9yjAio45j+4d4Uxebx5arEMJsP
I0XZCT1x7QT9pRcF0MTPashuEWGiXGsK1iuRGfH6ra3Fu4l29IhWVgiD2emwohQk
67o4D9ctVzz4kvqg6JiQCk0fXg6eL+7ryF4YHzKPHb/G0JXs0Jmb09rkvCnJCj1d
MZQesXX6Y6Mqm0QcVPTuWsohrqp3M53by0AlMSLnazvrfdj3nQ7bH00Dm6VXS/yX
mkOzQzB3Ec6vRVhnEv9UndzWD6lng3Auz7y93reN5RNYd1aX7aScIc4yKcezGDrz
sJ0Z0EGfAvRTSaOjd7Ka4xfmUMukAcQWHrDidhrnpuAwIYRtbSbSO9wBhcLvk7qq
uhsABi0jhT+Y6AzoI61Q
=1AEH
-----END PGP SIGNATURE-----

JeremyRand:
Since 185765198887d3610258586ab22f0d0c6438eaee is merged to master branch, we can close this PR now, right?

luccioman:
Yes @JeremyRand it was time to close this.

PR #97, opened 2016-12-22
Added entrypoint script for docker to catch SIGTERM

m0se:
The startYaCy.sh script does not catch SIGTERM from `docker stop` so the instance is killed after the `docker stop` command times out. 

The entrypoint script invokes stopYaCy.sh when it receives a SIGTERM and shuts down YaCy in a clean fashion.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -52,5 +52,7 @@ VOLUME ["/opt/yacy_search_server/DATA"]
 # Next commands run as yacy as non-root user for improved security
 USER yacy
 
-# Start yacy in debug mode (-d) to display console logs and to wait for yacy process
-CMD sh /opt/yacy_search_server/startYACY.sh -d
+
+# Add and set entrypoint script to start and stop YaCy 
+ADD entrypoint.sh /bin/entrypoint.sh
+ENTRYPOINT ["/bin/entrypoint.sh"]
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+trap "{ 
+echo Exit TRAP snapped
+
+/opt/yacy_search_server/stopYACY.sh
+
+exit 0
+
+}" SIGTERM
+
+/opt/yacy_search_server/startYACY.sh -d &
+
+wait


luccioman:
Hi @m0se thank your for your proposal. 
Did you also checked if a more generic solution could be applied? 
I mean, when starting YaCy without docker (either with `startYaCy.sh -d` or directly with `java` command) SIGTERM signal seems to be catched by YaCy's main [Shutdown hook](https://github.com/yacy/yacy_search_server/blob/master/source/net/yacy/yacy.java#L799), but the next shutdown steps are not performed... It would be great if SIGTERM could trigger a proper termination on all kind of deployments.

m0se:
I think a more generic solution would be to add a trap to the `startYaCy.sh` script. 
When sending a SIGTERM to the pid of the running `startYaCy.sh -d` process the script is terminated, but the java process is not shutdown. When the SIGTERM is sent to the pid of the java process then it quits (not sure if it's shuoting down cleanly, since it happens very fast in comparison to invoking the stopYaCy.sh).

With a trap in the `startYaCy.sh` script, actions could be defined (like calling the stop-script or send a SIGTERM to the java process) before the script itself terminates. In the docker context this is very important, since `docker stop` only sends a SIGTERM to the script that was startet with `docker run` via. CMD or ENTRYPOINT. So the only way to have multible shutdown / cleanup actions is trapping the SIGTERM in this ENTRYPOINT/CMD script.

It might would makes sense to have this in the `startYaCy.sh` script, I just don't wanted to mess with the original script just for docker and I first would have to look into how traps with `/bin/sh` work, I normally use `/bin/bash` for scripting. But if there is a interest in having this in the main script I can look into it.

luccioman:
Hi @m0se , in the meantime I [fixed](https://github.com/yacy/yacy_search_server/commit/1df558a6c6c86b96dd87987f7f8b87ad12b901c4) YaCy shutdown hook to properly terminate the Java process on SIGTERM signal.
I just experimented with using `exec` in startYaCy.sh when -d option is set, and by using the "exec form" of the docker CMD (i.e. CMD ["/bin/sh","/opt/yacy_search_server/startYACY.sh","-d"]) and this way `docker stop` seems now to properly stop YaCy server. A bit more testing is needed before committing, but don't you think this way is even more generic (no new docker specific script needed) and should be preferred?

m0se:
Hey @luccioman, I completely agree, this would be a more elegant solution. I tried to use the "exec form" you mentioned, but I assume you also did some changes in the `startYaCy.sh`, as I still have the same issue when using `CMD ["/bin/sh","/opt/yacy_search_server/startYACY.sh","-d"]`: the script terminates and the Java process is forcefully killed after timeout.
I confirmed this by doing it manually, sending first a SIGTERM to the script and then sending a additional SIGTERM to the still running Java process. After that the Java process started shutting down with your newly added stdout:
```
SHUTDOWN Starting shutdown via shutdown hook.
SWITCHBOARD Performing shutdown steps...
SWITCHBOARD Shutdown steps terminated.
```
So that part works now, but I think the signal send to the script (as it is now in master) still doesn't makes it to the java process.

luccioman:
Yes for the signal to be properly propagated, I had to change both the Dockerfile CMD and the startYaCy.sh script. I pushed [my modifications](https://github.com/yacy/yacy_search_server/commit/1eafa7bfaf6b535123894e9267c3b7e670981686) and rebuilt the images on Docker Hub.

Tests were successful (using docker stop and checking no regression on a standard Linux deployment) but of course your feedback will also be welcome.

luccioman:
Hi @m0se , can we close this PR now?

m0se:
Sorry for the delay, I just pulled the latest debian based yacy docker (9aea8d194c3d) and gave it a try. Works like a charm! When hitting `docker stop` the shutdown hook is triggert.
Only thing I noticed is that the container does exit with `Exited (143)` and not `Exited (0)`, but for my deployment this doesn't matter.
Anyhow, thanks for the support!

PR #100, opened 2017-01-01
upgrades to jetty 9.4.0.v20161208 and solr 6.3.0

automenta:
it seems only minor changes were needed for both dependencies.  as far as i can tell the unit tests are not affected, although the same 5 did fail before and after these changes, i think this must be normal.  these tests were run on jdk9.


automenta:
btw sorry for the extra IDE .xml files that were commited, these should not have been part of the commit

automenta:
also i disabled sevenzip file type support. this needs to be re-enabled

automenta:
and there seems to be a bug with jetty when i actually try to run the web interface, ill see if i can fix this in next commit

PR #101, opened 2017-01-01
upgrades to jetty 9.4.0.v20161208 and solr 6.3.0

automenta:
it seems only minor changes were needed for both dependencies. as far as i can tell the unit tests are not affected, although the same 5 did fail before and after these changes, i think this must be normal. these tests were run on jdk9.

the second commit adds the upgraded dependencies to the lib/ which was needed for ant and startYACY script. seems to work


luccioman:
Hi @automenta , I suppose you are aware that this upgrade implies no more support of Java 7. This is planned, but personally  I am not convinced it is already the time. For example the current stable Debian (Jessie) is still provided with openjdk 7 as the default Java implementation. And I am not sure it is a good idea to force Debian users to install openjdk 8 backport package.
But the opinion of other YaCy developers and users would be interesting.

By the way, thank you for testing this upgrade.

Best regards

automenta:
i like jdk9 https://jdk9.java.net/download/ unfortunately jetty gave an NPE in a WebAppContext static block that could not be overridden, and this alone forced my use of jdk8.  to continue i would replace jetty with undertow, particularly for its efficient websockets https://github.com/undertow-io/undertow

luccioman:
Ok @automenta but as far as I know most of YaCy users are currently rather using JDK7 or JDK8, so to my mind it is not the time to already force everyone to quit JDK7.
Switching to another web server implementation won't be a lightweight refactoring, and even if the new chosen server could be itself more efficient, it doesn't mean YaCy will automatically have performance improvements... To my mind if we would like to use WebSocket features we should first do it with Jetty (but I agree I don't know if Jetty 9.2.x as sufficient WebSocket support).

automenta:
https://cwiki.apache.org/confluence/display/solr/Major+Changes+from+Solr+5+to+Solr+6

automenta:
yes, maybe i should revert that

reger24:
Hi @automenta ,
your pull request hast some meanwhile some conflicts BUT helped a lot to upgrade to Jetty 9.4, see
https://github.com/yacy/yacy_search_server/commit/e6e20dab5279900a9929dd6935cac97b48bef249

Thanks for your work, input and inspiration.

PR #104, opened 2017-01-23
(WIP) Javascript Re-Sorting

JeremyRand:
(WIP) Javascript Re-Sorting

This PR implements re-sorting in Javascript.  See the discussion in https://github.com/yacy/yacy_search_server/pull/91 for background on why this is useful.  In my testing, the relevance of results is much, much improved with this PR applied.

Testing and feedback are encouraged; the following known tasks remain before it could be merged.

- [ ] Add a UI setting to control whether Javascript re-sorting is enabled.  Right now, enabling it requires changing a pref manually.  I wasn't sure what page to put the UI setting on; feel free to suggest a place.
- [ ] Remove commit c1c97214f4917da478232584e6526a85a95fe829 ; it's only present to make debugging more convenient.  This commit could be replaced by https://github.com/yacy/yacy_search_server/pull/78 for people who want it.

### DIFF
--- a/defaults/yacy.init
+++ b/defaults/yacy.init
@@ -907,6 +907,11 @@ search.excludehosth=
 # the cases of nocache, iffresh and ifexist causes an index deletion
 search.verify.delete = true
 
+# If enabled, the results are sorted in the browser using Javascript.
+# This usually improves ranking accuracy, but doesn't work well for users
+# who have Javascript disabled, are using screen readers, or are on slow computers.
+search.jsresort = false
+
 # remote search details
 remotesearch.maxcount = 10
 remotesearch.maxtime = 3000
--- a/htroot/js/yacysearch.js
+++ b/htroot/js/yacysearch.js
@@ -188,7 +188,7 @@ function statistics(offset, itemscount, itemsperpage, totalcount, localIndexCoun
 	  progresseBarElement.setAttribute('style',"width:" + percent + "%");
   }
   var resnavElement = document.getElementById("resNav");
-  if (resnavElement != null) {
+  if (resnavElement != null && !jsResort) {
 	  resnavElement.innerHTML = renderPaginationButtons(offsetIntValue, itemsperpageIntValue, totalcountIntValue, navurlbase, localQuery, jsResort);
   }
 }
--- a/htroot/js/yacysearch.js
+++ b/htroot/js/yacysearch.js
@@ -42,9 +42,12 @@ function fadeOutBar() {
 }
 
 /**
+ * @param offset item number to start with
+ * @param itemsperpage count of items requested per page
+ * @param totalcount count of items available from YaCy node for this query
  * @returns pagination buttons
  */
-function renderPaginationButtons(offset, itemsperpage, totalcount, navurlbase, localQuery) {
+function renderPaginationButtons(offset, itemsperpage, totalcount, navurlbase, localQuery, jsResort) {
 	var resnav = "<ul class=\"pagination\">";
 	var thispage = Math.floor(offset / itemsperpage);
 	var firstPage = thispage - (thispage % 10);
@@ -52,7 +55,11 @@ function renderPaginationButtons(offset, itemsperpage, totalcount, navurlbase, l
 		resnav += "<li class=\"disabled\"><a title=\"Previous page\" href=\"#\">&laquo;</a></li>";
 	} else {
 	 	resnav += "<li><a id=\"prevpage\" title=\"Previous page\" accesskey=\"p\" href=\"";
-	    resnav += (navurlbase + "&amp;startRecord=" + ((thispage - 1) * itemsperpage));
+	 	if (jsResort) {
+			resnav += ("javascript:numberedPage(" + (thispage - 1) + ");");
+		} else {
+			resnav += (navurlbase + "&amp;startRecord=" + ((thispage - 1) * itemsperpage));
+		}
 	  	resnav += "\">&laquo;</a></li>";
 	}
 	
@@ -66,7 +73,11 @@ function renderPaginationButtons(offset, itemsperpage, totalcount, navurlbase, l
 	       resnav += "</a></li>";
 	    } else {
 	       resnav += "<li><a href=\"";
-	       resnav += (navurlbase + "&amp;startRecord=" + (i * itemsperpage));
+	       if (jsResort) {
+                       resnav += ("javascript:numberedPage(" + (i) + ");");
+	       } else {
+		       resnav += (navurlbase + "&amp;startRecord=" + (i * itemsperpage));
+               }
 	       resnav += "\">" + (i + 1) + "</a></li>";
 	    }
 	}
@@ -74,7 +85,11 @@ function renderPaginationButtons(offset, itemsperpage, totalcount, navurlbase, l
 		resnav += "<li class=\"disabled\"><a href=\"#\" title=\"Next page\">&raquo;</a></li>";
 	} else {
 	    resnav += "<li><a id=\"nextpage\" title=\"Next page\" accesskey=\"n\" href=\"";
-	    resnav += (navurlbase + "&amp;startRecord=" + ((thispage + 1) * itemsperpage));
+	    if (jsResort) {
+		resnav += ("javascript:numberedPage(" + (thispage + 1) + ");");
+	    } else {
+		resnav += (navurlbase + "&amp;startRecord=" + ((thispage + 1) * itemsperpage));
+	    }
 	    resnav += "\">&raquo;</a>";
 	}
 	resnav += "</ul>";
@@ -95,7 +110,7 @@ function parseFormattedInt(strIntValue) {
 	return intValue;
 }
 
-function statistics(offset, itemscount, itemsperpage, totalcount, localIndexCount, remoteIndexCount, remotePeerCount, navurlbase, localQuery, feedRunning) {
+function statistics(offset, itemscount, itemsperpage, totalcount, localIndexCount, remoteIndexCount, remotePeerCount, navurlbase, localQuery, feedRunning, jsResort) {
   var totalcountIntValue = parseFormattedInt(totalcount);
   var offsetIntValue = parseFormattedInt(offset);
   var itemscountIntValue = parseFormattedInt(itemscount);
@@ -174,7 +189,7 @@ function statistics(offset, itemscount, itemsperpage, totalcount, localIndexCoun
   }
   var resnavElement = document.getElementById("resNav");
   if (resnavElement != null) {
-	  resnavElement.innerHTML = renderPaginationButtons(offsetIntValue, itemsperpageIntValue, totalcountIntValue, navurlbase, localQuery);
+	  resnavElement.innerHTML = renderPaginationButtons(offsetIntValue, itemsperpageIntValue, totalcountIntValue, navurlbase, localQuery, jsResort);
   }
 }
 
--- a/htroot/js/yacysort.js
+++ b/htroot/js/yacysort.js
@@ -28,8 +28,20 @@ var displayPage = function() {
     }
   });
 
-  $("#offset").html($("#resultscontainer .searchresults.earlierpage").length + 1);
-  $("#itemscount").html($("#resultscontainer .searchresults.earlierpage").length + $("#resultscontainer .searchresults.currentpage").length);
+  // TODO: The following statistical displays could maybe be moved to the latestinfo() call.
+
+  var offset = $("#resultscontainer .searchresults.earlierpage").length + 1;
+  var itemscount = $("#resultscontainer .searchresults.earlierpage").length + $("#resultscontainer .searchresults.currentpage").length;
+
+  // TODO: This seems to often be smaller than the "totalcount" that statistics() ends up with.  Why is that?
+  var totalcount = $("#resultscontainer .searchresults").length;
+
+  $("#offset").html(offset);
+  $("#itemscount").html(itemscount);
+
+  $("#resNav").html(renderPaginationButtons(offset, requestedResults, totalcount, null, theLocalQuery, true));
+
+  //latestinfo();
 
   console.log("Showing results " + ($("#resultscontainer .searchresults.earlierpage").length + 1) + " - " + ($("#resultscontainer .searchresults.earlierpage").length + requestedResults) + " out of " + $("#resultscontainer .searchresults").length + "; notEarlierPage = " + $("#resultscontainer .searchresults:not(.earlierpage)").length);
 };
--- a/htroot/js/yacysort.js
+++ b/htroot/js/yacysort.js
@@ -72,6 +72,163 @@ var laterPage = function() {
   displayPage();
 };
 
+var processSidebarNavProtocols = function(navProtocolsOld, navProtocolsNew) {
+  navProtocolsOld.find(".btn-group-xs").each( function(index, oldProtocol) {
+    var protocolId = $(oldProtocol).attr("id");
+
+    var newProtocol = navProtocolsNew.find("#" + protocolId);
+
+    // Check whether the protocol has been removed in the new sidebar.
+    if (newProtocol.length === 0) {
+      console.log("Deleting nav-protocol...");
+      $(oldProtocol).hide(1000);
+    }
+  } );
+
+  navProtocolsNew.find(".btn-group-xs").each( function(index, newProtocol) {
+    var protocolId = $(newProtocol).attr("id");
+
+    var oldProtocol = navProtocolsOld.find("#" + protocolId);
+
+    // Check whether the protocol exists in both the new and old sidebar
+    if (oldProtocol.length === 1) {
+      // Replace the HTML.
+      // TODO: Look into smoother animations.
+      $(oldProtocol).html($(newProtocol).html()).show(1000);
+    }
+
+    // Check whether the protocol has been added in the new sidebar.
+    if (oldProtocol.length === 0) {
+      // We need to insert the protocol in the right position.
+
+      // TODO: Insert in the correct position instead of the end.
+      $(newProtocol).hide();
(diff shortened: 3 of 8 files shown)


JeremyRand:
@luccioman Would you be interested in testing this out?

luccioman:
Hi @JeremyRand yes I am interested, I just wanted to have sufficient time to test your proposal and then give you a relevant feedback. This should be soon.

JeremyRand:
Hash: SHA512

luccioman:
> Hi @JeremyRand yes I am interested, I just wanted to have
> sufficient time to test your proposal and then give you a relevant
> feedback. This should be soon.

Okay, no worries.



-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJYkEBQAAoJELPy0WV4bWVwYPwP/j/CjmKZu029vUY93QwsMTJO
GH7LOh+JJp/KesmLCXoF1jGXim6PgKJTjs6GidDi2XXVcUz89cXMOOKqlNukWL+q
xFfOoeM3VwlQVy7a4g57rAORnWEKEEimkcygdIbOlv3GrgN4TyNBRQ4v3K2rlvtS
8KfDpsHPx9kgU8+4vQf0E1rMF4Wh79qbK6zf2NFLNeDUuGZkB0TWdGPUoymWoEmy
CP76G6RRxyEaL31Xi2UBWGulVeQYkI2g7RHYN4TDV5fgW/cnMF2UxbWThKnyYlaJ
zrUaRZOTcrwu/9uZXARfysnQ3ir91B2nEmyGTFMxkFnG5QeZX/+g8bCL1W12wCFD
xEAJNDuRdYUlykWZfcGCXXW56FRjT08lhyT2O0X65OCvM6iHG7hH7YEu4rZapDAP
eAFPioJ0Q9oDce+nMjaull1qnGx9d96aZ+jr042p9a4zc6quCqcALxrGDYsky3hk
1r+8ed1t+47aeHFTL1uwOBUskL6tqVBLPwynUEt9B9XvfsmB1fr2JDvTChHDE3Mf
2qnP5ZHiNPahTOTV5AGsw08G4IjX5/5W3msyPd/mt6pS0aPIT/Gce9VVDxhgkZbO
tsfx7nIau08eacoG8hp4AfUx4rUSeYW1L/edUNDzWHPejbEpKdBf8E2V+RmTtCpm
OOi9sU/alxRyijZIhOCh
=y0z5
-----END PGP SIGNATURE-----

luccioman:
Hi @JeremyRand I will continue testing your proposal in the coming days, but I can already confirm it is working quite well. I tested with two freshly installed YaCy peers face to face, one with js resorting and the other without. JavaScript resorting appears to be quite useful and to my own taste lower rank results disappearing is not so disturbing.
A few remarks however :
 - facets are no more displayed. Is it intentional?
 - using the same pagination buttons as usual and allowing to go directly to a specific page would be great
 - even when all remote results have been fetched, it looks like the browser continue to regularly request indefinitely the server ( GET XHR http://host/yacysearchitem.html traces in the browser console)

By the way thanks for you work.

JeremyRand:
Hash: SHA512

luccioman:
> Hi @JeremyRand I will continue testing your proposal in the coming
> days, but I can already confirm it is working quite well. I tested
> with two freshly installed YaCy peers face to face, one with js
> resorting and the other without. JavaScript resorting appears to be
> quite useful and to my own taste lower rank results disappearing is
> not so disturbing.

Great, thanks for reviewing -- glad to hear that it works well for you.

> - facets are no more displayed. Is it intentional?

Not intentional.  I rarely use facets so I didn't notice.  I'll
investigate this.

> - using the same pagination buttons as usual and allowing to go
> directly to a specific page would be great

I think I can do this, subject to 2 caveats:

1. The pagination buttons will be handled by Javascript rather than
standard HTTP links, and won't be accessible via permalinks.
2. If new results are still arriving when the next pagination button
is clicked, some of the currently displayed results might show up on
the next page; similarly, some results might get skipped when the
previous pagination button is clicked; and clicking on the same page
number over and over might cause the results shown to change slightly
each time it's clicked.  None of these issues will occur if new
results have ceased arriving when the pagination buttons are clicked.

Both of these are because the pages are defined by a floating-point
maximum score rather than an integer page number, and the maximum
score for each page will increase slightly as more relevant results
arrive, so it's not possible to deterministically predict what the
maximum score for a given page will be.

If these caveats are okay with you, I'm happy to try implementing this.

> - even when all remote results have been fetched, it looks like the
> browser continue to regularly request indefinitely the server ( GET
> XHR http://host/yacysearchitem.html traces in the browser console)

Good point.  I wasn't certain how to disambiguate between "we've
fetched all the results the server has right now, but more may arrive
soon" versus "the server has stopped listening for new results, so we
can stop asking the server".  Does yacysearchitem.html return any info
that would tell the browser this?  Or does YaCy use a constant-length
timeout after which it stops listening for new results (in which case
it may be simpler to just have the Javascript stop asking after that
period has elapsed)?


-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJYk2LIAAoJELPy0WV4bWVwmSkQALx4U6pBUW0qAf10u+KfZNUT
u8nzObTcOz6aqelHdY8HedEETPq8ficXfpWt6/XmS9aWR1ts0TOnDdcH/esSVHIz
MFyl+S2m3U1EaO8EBR0+JesaN0llmT7AWQnZS7fQVdlXJHaNGPrXu8ZpW+9sd4oX
V+hkzIk9M93lX2j9d3klwxVUsP3nQ8hAYlfwOlnNXbZIukOqtq7Mw5/w/MFYlhzz
m9VfBD5ZrwXsVEmn9R3I0QNRvnMq/tQnJggJ7+CLK+3LzoL/eN9yMtr17RKI3vBR
ZwY6eYv7lutTnAvFOIRU/xyqspJnVUs57NUm1QmOx9C/xY7WD4F7AbCcLr+5Ea3H
csTUJ9GIvXnuSB7JNbQQK7QU9Ltj8UK/Uz0A+MvZvSdv2dhIIXGDjot7iw5b0Zif
mE5BWZY4B4UUS3bQJmOikhse9ps8OjjvsV7lwNKNJRv1oag8Q4eoIF/LC81lqva0
q+CwkaR3Vx5L3OJgwjWo/FzW9kgrNbwfUTfSWi4o+Gy3UBEMifc45DpZbUIfO2FO
wRuDxpB6wWqS6CR1rN1vqn6P3UQXZ6M+p8igDzrgLQFG9KeBQVOJ7/THvyuoCTTO
ZhpgdbS2xqolQH/CwSQguOBXhgbpQVsChENpPEEoF+BRg5ts5E9Wfw97z1bT4IZL
H5P1cY2mOl6Qjiiv2siW
=ll73
-----END PGP SIGNATURE-----

luccioman:
> If these caveats are okay with you, I'm happy to try implementing this.

To my mind JavaScript buttons should not be a problem here as at this point we have enabled JavaScript resorting. About a part of non-determinism : I believe we should neither be afraid by it considering the current YaCy P2P mode search behavior ;)

About retrieving whether 

> the server has stopped listening for new results, so we can stop asking the server


Yes I dug again the code to find this but it looks like this information is missing. Maybe it would be a good idea to render it in the yacysearchlatestinfo.json API. As far as I know, it would involve checking the running status of the different results feeding threads (SearchEvent#localsolrsearch, rwiProcess, primarySearchThreadsL, nodeSearchThreads and secondarySearchThreads).

JeremyRand:
Hash: SHA512

JeremyRand:
> luccioman:
>> - facets are no more displayed. Is it intentional?
> 
> Not intentional.  I rarely use facets so I didn't notice.  I'll 
> investigate this.

I just pushed a commit that mostly fixes this.  Let me know if you run
into any issues with it.

Cheers!

-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJY4bZRAAoJELPy0WV4bWVwEX8QAIVsZL5gFeQpuZ5s2MeGakPb
D3VktKddB4Ek6ii1UtZpTtBQAFiaZvkMsxSua2cIyIb+PT1HgAorII6w+jVs0+09
RIEjkq4tJAyRcMU3GYWFFuV0wESpABiHAghJZFgMhFrlHwnVFLT6Q7/klKyEg9Sf
DjeI8f+9WTOPGhtowlor2mnsZwQ8wrv6kiP8I6V9Yu2xfafsKaH2M5ktFaYFIL6E
YPUJWhJPYVLmMSNOFFE+LxC/u2kZyKZy83/6t//N9FmJ+mySAseKI3jj+8DjboDJ
8EYt5byPPH5IG0AYSQTNNLDRSsV+QNCD6FsvLudA5yAn4Itp/C+nClOSpoDGxMLV
tVOPLjBxNMG0mex1esdQeh0l+imztgZksNpqcY2VX8SGvhxHxqphL0hxTVvHTWIR
i9EtiY2hUs9UACs+PdRPySdy6+MEeaVlQibCvpDNc8cK7Ncce4lFNSoq41QO20oZ
Jp5j7dbRgv1/NiUDwumkLbt8XdT2Sckn4f8wwkTw/B3qsHSi7hvVeHEY+4fF7m+h
eqCplDytjBBQgkOcpkcyisapqm8mQPXd/p1T/7N96PjbsA38GCRrKcYKF9iTkkOh
YolSJogRAJ+jf6/XXPpF8Lq29kRQA/76xfSK3ZphD8wun8HREPF/1m1I6fs3n52g
P85gcp4Ro7OFdR+gTU13
=AgSZ
-----END PGP SIGNATURE-----

luccioman:
Hi @gardner, I continue on your last comments posted on issue #91 and agree it would be the time to have a deeper look at this PR. 

For your information, on YaCy main's branch there is now already a possible alternative way to communicate rank : in the "Debug/Analysis Settings", you can tick the "Show search results scores". Then the raw ranking score value is displayed for each result.

There is also now an visual indicator about the status of remote results background fetching. If we add the possibility to automatically refresh the current page results when fetching is finished, this could meet your third mode proposal :
> Update UI after a certain condition

By the way realtime re-sorting may still sound interesting for some users. Some TODOs remain in the yacysort.js of this Pull Request, so I guess there is some more testing and code review to perform before merging. If you have any feedback you are welcome.

JeremyRand:
Hash: SHA512

For the record, I do plan to finish this PR, I've just been incredibly
busy with other things and haven't had time to touch this in a while.
 I'll try to find time to come back to this soon.

-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJZm8tLAAoJELPy0WV4bWVwJbUQALXfeGJvsAP61ianra2ONXd4
WdqJcXOLY8u/Y2hHmOeuGfadaKHWTMqsdGY5bYpDdg3Kb7K3V8BvxunvQG5DgIwz
lz+c40cFwE0IXqsCGuhkUk4145dtitdthUwGCP+wVk9eW0aQQQuTp3/KuACFwfoe
qAaLxsufhfvsSG6naBrlxUNH2XgKCmlTEfUxvUDK+L16aJ29Ck+m3sWIKRK/3Few
YoDbzdKMBLPtedtaapC4DMrPLe+yO1qLEdrP4U1+ifgh6lQJp2k9GTmnJ5nkMEb2
f27BhkUgaaekNTGlMwKB8W0QcWXR0f4AYG+S1Tqm6s6m6dBKSwkSbhYjDvYBrsad
cs2KvfOslCRk8oIvqjQi4Er2mJL+Fb0Eu9H+jHmwLz7sZbpsnJtPpdhUgmmfh1SI
RBvNj0dis0XYmkSVfDlJSaueR4yzY1wUob1lvoerKZKAcLsTdJLfAbvZs045bfp+
OdaYvJlju08U9QTWQJidRl0qRRdXb7tQpnFwfrrZWAZJXmY/zaeTN5v+99/wpe6x
Q0DoSQcBtg+SjSciCxbVSrfIr6TBn1VvBoSBjHG6FFFVZzZ5VXBkJIaOEH368WBh
M3gGpSM6ujG4rCoSMVr34zV4Sr8JjVxqH5LVMiG9xsp+nbYwd9VybW565W/Ty8iG
8DxcsXX+WvQls2uwuH/i
=9rqY
-----END PGP SIGNATURE-----

JeremyRand:
@luccioman:

> using the same pagination buttons as usual and allowing to go directly to a specific page would be great

This should be working now as of 59e16002dfda55f37185b6313445a73570eb0c99 .

Next I'll try to get the merge conflicts resolved, hopefully that won't be incredibly hard.

JeremyRand:
> Next I'll try to get the merge conflicts resolved, hopefully that won't be incredibly hard.

Merge conflicts are now resolved.  Would be awesome if anyone wants to review what I have so far.

luccioman:
Hi @JeremyRand I will try to review and test your latest work this week. I also just worked on the subject from the server-side perspective (see commit a1a0515 ) and luckily did not introduced again more merge conflicts :)

Scarfmonster:
I did some quick testing, the results really do appear to be more consistent and relevant. The biggest increase in quality seems to come from big peers which possibly hold a lot of relevant results, but because of that may be slower to respond. The scrolling isn't necessarily a bad thing, as it gives the user something to look at, and sometimes you get what you came for before all the results have been fetched, meaning shorter wait times.

I don't really see the use of JavaScript as a problem since the current search page uses it as well. Although a lightweight, non-js option may be a nice thing to have in general. 

Mobile browsers struggle a bit with that amount of moving things though, it's not anything severe, but causes visible animation stuttering. Maybe there should be a configurable setting how often updated results should be fetched and/or displayed? Fast moving objects could also have a negative impact on people with epilepsy or cause discomfort, and the refresh speed setting is a simple way not to force them to disable the feature.

Sometimes the sidebar keeps updating a long time after all the results have been displayed. Most visible are the keywords, but other things are updating too. Possibly it could have its own indicator that internal counts are still updating even after all remote results have been fetched?

If the “shallow crawl on all displayed search results” heuristic is enabled, it seems to treat all remote search results as displayed, filling the crawl queue with thousands of entries. It can be both desired and undesired thing though. I am wondering if instead of “all displayed” it would be better to use something like “X top ranked results”. This isn't really a problem with this PR though, just how that particular feature works.

JeremyRand:
Hello @Scarfmonster!  Thanks for testing.  I'll reply to your other comments a little bit later, but regarding this:

> Mobile browsers struggle a bit with that amount of moving things though, it's not anything severe, but causes visible animation stuttering.

I just ran the sorting code through Firefox's profiler, and it might not be the animation that was causing performance issues on slow devices, but instead some poorly optimized selectors.  I've added a commit that improves performance in Firefox's profiler by quite a bit for me.  Any chance you could test again and see if it helps on your mobile device too?

(Also, I notice that there are again some merge conflicts, so I'll try to fix those now.)

JeremyRand:
Merge conflicts are fixed.

luccioman:
Thanks guys for testing and maintaining this PR up to date. 
Sorry for the supplementary merges required... Before introducing this to the main branch, I just wanted first to check and improve/fix points in that code base area which can be handled in a reasonable amount of time the.
For now I still have some work in progress related to navigators counts (facets), but I think it is going into the right direction.

Scarfmonster:
@JeremyRand

I'm without access to a computer right now, so I can't test this properly. From what I could see, it is a bit faster but still has trouble when a new element appears at the very top and it has to push everything down. I can't check this right now but, if my memory is right, it has to do with how jQuery's show and hide behaves. I will try to replace it with css animations on the weekend.

Scarfmonster:
I attempted replacing the JS animations with CSS animations and it appears to work more smoothly. It is a bit quirky though, as it limits the height of a single result to 500px. I don't believe anybody will encounter a single result surpassing this though. Scarfmonster@634f52fefcee0ceecff8cf09add6fe6cb80615b4

luccioman:
Hi @Scarfmonster I tested with the custom CSS animations you pushed on your branch, and I also obtain definitely smoother results with less CPU consumption on desktop browsers. 
I also tried to use the JQuery animations with an upgrade to version 3.2.1, and did some profiling : the frame rate is improved, but custom CSS animations still appears to consume less CPU.

However, I do not get why you have to limit the height of results to 500px within your CSS animation frames. Is it to prevent some cross browsers bad behavior? Using only the scale transform from 0 to 1 (or the reverse) appears to be working fine in my tests...

Scarfmonster:
@luccioman
It's because using scale transform doesn't scale the bounding box. If I didn't use max-height,  results would jump, instantly creating a gap in which a new result would then upscale. The problem with max-height (and height), though, is that there is no way to animate it to a computed height, so a value has to be used which is big enough to fit all results in normal circumstances.

luccioman:
Thanks @Scarfmonster for the clarification. Playing a bit more with CSS animations i also realized that you can not use the _display_ attribute in transitions or animations, so again, using the _max-height_ as as an alternative to collapse the bounding box makes sense.

Scarfmonster:
Thank you @luccioman for just making me realise that I somehow left display in my animations, while explicitly using overflow hidden to not use the display property... Anyway, I removed it since it can potentially break animations (and definitely breaks transitions). I also reset the max-height at the very end of the animation to none. That way, even in the very unlikely event of the search result being higher than 500px, it will display it fully. Again, it will have the jumping issue, but it's better than not being able to fully see the result.
Scarfmonster@2af011243fa62467acbdf71296958640c218be39

luccioman:
Hi, for information I did some complementary work on this subject on my own branch (https://github.com/luccioman/yacy_search_server/tree/jsResort) to prepare an eventual merge on the master.
Some TODOs remain and more cross browser testing is required to my mind, but it's in progress.

luccioman:
The feature is now on the master branch, and can be activated in the ConfigPortal_p.html admin page for anyone already interested in using it.

For now I restricted its use to authenticated (or from localhost) users only as it still adds quite a high load to the server when intensively searching. But performance improvements are to come.

Also facets navigators are not yet all refreshed in realtime adn with animations, but it's coming.

By the way, to my mind it is already really an improvement for search results ranking accuracy. Thanks @JeremyRand for the proposal.

PR #105, opened 2017-01-26
Update README.md - removes deprecated URL

ivar:
removes url that now redirects to German forums (already listed)


Orbiter:
Thank you!

PR #110, opened 2017-02-23
Fixing some typos

goofy-bz:
up to line #1000 only


luccioman:
Thank you @goofy-bz !

PR #128, opened 2017-09-30
Prevent Integer Overflow in ChartPlotter

otteresk:
For nodes with indexes having more than 10 million entries the calculation of the y coordinates generates an integer overflow in chartDot and chartLine. Using long for the multiplication fixes this.


luccioman:
Thanks @otteresk , yes the chart is working better with your fix. 
However I found and fixed some additional integer overflows in the ProfilingGraph class (on pixelperscale parameter calculation).
Just to be sure, on which graph exactly did you encountered the issue? PerformanceGraph.png or NetworkHistory.png?

otteresk:
@luccioman , thanks for including. I found the overflow in NetworkHistory.png. E.g.my node: https://v22014012007616699.yourvserver.net:18043/NetworkHistory.png?columns=cR&scale=2000&maxtime=20000

luccioman:
Ok, as far as I tested your fix do the job for the NetworkHistory.png graph... up to Integer.MAX_VALUE documents on a peer (more than 2 billions). But I guess currently this is not really an issue as there are not much more documents on the whole YaCy freeworld network...

PR #137, opened 2017-11-10
ITA translation 

SebastianoPistore:
update italian translation.


luccioman:
Hi @SebastianoPistore that is great you updated the translation file for Italian. But it looks like there are also many German translations included in your version. Would you like to do some cleanup?

SebastianoPistore:
The translation is just beginning: the first thing I did was update the file at the last German version (the last Italian translation was at least two years). Then I started translating: I cannot open file with Poedit or OmegaT, so I edited it manually. But so the job is long and uncomfortable: I uploaded a first uncompleted version to update the code (translation is not yet ready to enable Italian for interface languages).
Do you think I should first delete all the strings in German?

luccioman:
This project has only one single master branch, so everything that is pushed on it has to be as clean and stable as possible. So to my mind, yes even if partial (which is also the case to a certain extent  for the other translation files), the Italian language translation file only has to contain relevant translations.

Also for a more comfortable translation process, did you notice that YaCy embeds a custom translation editor (at /Translator_p.html page, link available in the bottom of the /ConfigLanguage_p.html page)?

But if you would like to use your preferred external editor, YaCy also has the capability to generate and convert xliff files, that should be compatible with OmegaT for example. For that last possibility, it is possible that you need to play a little bit with Java code to launch the TranslationManager class in the proper way. If you are not at your ease with Java and it would be useful for you, I can check that last point to see if something can be done to make the xliff import/export operations a little bit more user-friendly.

SebastianoPistore:
I see the translation manager in YaCy, but I prefer to use my program to take advantage of TM.

I had not thought about launching to export the file: it's a good idea (I'll try it even if I'm not very good with Java).

167/5000
To sum up: Is it ok if I commit partial translations, if the translations that have not yet been made are left blank?
I went ahead (slowly) with the translation.

luccioman:
> To sum up: Is it ok if I commit partial translations, if the translations that have not yet been made are left blank?

Not really : if the lng file contains something like this `YaCy Network Access==`, the text will disappear from the translated page, which is not a good idea as it is preferable to have some untranslated text in English rather than nothing. So the proper way for partially translated pages can be either :
- to comment not yet translated entries (start with a '#' character, for example `#YaCy Network Access==`
- to put the same value as the original English text (for example `YaCy Network Access==YaCy Network Access`
- to include only translated entries

And of course testing that the finally translated page still works as fine as the original is preferable.

PR #149, opened 2017-12-12
fixed loading default settings in Performance Settings of Queues and Processes

Scre13:
load was populated with memprereq instead of loadprereq when using Double.parseDouble(getConfig(SwitchboardConstants.INDEX_DIST_LOADPREREQ, "4")) in deployThread()


luccioman:
Thank you @Scre13 for your careful reading and fix.
I would just add however that the case was quite unlikely to occur as it would have occurred only after a manual edition and removal of some `nnn__loadprereq=value` settings in the default settings file (yacy.init).

PR #151, opened 2017-12-25
Added automatic recrawling of older documents

Scre13:


luccioman:
Hello @Scre13 , thank you for your proposal,
but why not reuse the existing RecrawlBusyThread that can be started from the IndexReIndexMonitor_p.html page? What's more its old documents selection query can be customized.
To make it fully automated, it appears that the only thing missing would be to add an entry to the recorded API calls when using it, so it can then be scheduled using the Process Scheduler (/Table_API_p.html page).

One more remark : I am not sure of the interest of removing redirected URLs from the index directly in the HTTP loader :
- it may add unnecessary CPU load at this step
- redirected URLs are not really duplicates in the index, as not text is normally indexed, just a failreason_s such as "TEMPORARY_NETWORK_FAILURE cannot load: load error - CRAWLER Redirect of URL..."
- some users may prefer to keep a trace of redirected URLs in the index for their own custom usage

Scarfmonster:
@luccioman @Scre13 

` I am not sure of the interest of removing redirected URLs from the index directly in the HTTP loader`

Won't removing these links in the loader potentially cause them to be requested as many times as they appear on crawled pages? For example, if there is a link included on every page on a website, and it redirects somewhere, for every page it appears on, won't YaCy extract it, download it, check for redirect, remove it from index and then repeat this the next time the same URL appears, since it has no history of the URL existing?

luccioman:
@Scarfmonster , as far as I understand the current code base, and running/debugging again a few test crawls, prevention of repeated known redirections is currently rather done using the Web Cache : response of redirected URLs are eventually stored in the cache, so later requests on these same URLs, depending on the cache strategy, use the stored response from cache instead of following again redirection(s) to fetch content.

Scre13:
@luccioman, i wrote a new class, to keep code seperate and not to interfere with the existing RecrawlBusyThread, as I have some more ideas to improve it in the future.

@all 
The reason for removing the original URL in HTTPLoader.java is the result of my tests.
As fresh_date_dt and load_date_dt are not updated on the original URL I decided to remove original URLs and use only the redicted URLs.
As I use fresh_date_dt and load_date_dt in a solr query to find new URLs to be recrawled, this would cause a loop. 
BTW: I think the existing RecrawlBusyThread has the same problem.

luccioman:
Thanks @Scre13 for your explanations. Of course if the existing recrawl task can lead to infinite loop this should be fixed. But personally I would prefer a solution where existing indexed URLs do not need to removed on any HTTP load.

On your proposal in general also, to my mind it is valuable to provide an easy way to automate recrawl of old index entries, but we should be careful before imposing this kind of background task to every peer. Especially for Intranet or Webportal modes, I am not sure this is necessarily a good idea.

PR #219, opened 2018-08-17
misc cleanup, and unverified but possibly successful attempts at performance improvement

automenta:
sorry this commit is a bit messy. im not suggesting to pull it all in but only the following files which i'll briefly explain what is different:

source/net/yacy/cora/order/Base64Order.java

misc work here, including extracting main( ) to a unit test.  mostly simple changes to the data structure access patterns that i thought could enhance performance:
source/net/yacy/cora/sorting/AbstractScoreMap.java
source/net/yacy/cora/sorting/ConcurrentScoreMap.java
source/net/yacy/cora/sorting/OrderedScoreMap.java
test/java/net/yacy/cora/sorting/OrderedScoreMapTest.java

same for the ARC files.  just some updates to the way Map is accessed.  extracted a similar main() unit test also which is applied to both Hash and ConcurrentARC
source/net/yacy/cora/storage/ConcurrentARC.java
source/net/yacy/cora/storage/HashARC.java
source/net/yacy/cora/storage/SimpleARC.java
test/java/net/yacy/cora/storage/ARCTest.java

i forget what changed here but it should be minor:
source/net/yacy/kelondro/data/word/Word.java

i replaced collection generation with an iterator.  not sure if this breaks anything.  the other change here is one application of a CountDownLatch instead of the Thread.join waiting.  a similar pattern can probably be applied elsewhere.
source/net/yacy/search/schema/CollectionConfiguration.java

all unit tests except a scraper test which isnt affected continue to pass.  everything seems to run smoothly when i launch the server.


luccioman:
Thank you for sharing your work on YaCy @automenta .
But yes, you would do a favor to anyone interested in reviewing and testing your contributions by creating distinct Pull Requests with a clearly defined goal in each one ;)

And about modifications with expected performance improvements but without any measurements confirming that... To my humble opinion, we should be careful with that kind of approach, as you can easily obtain the opposite result than what is desired, for example regressions or performance reduction.

By the way, I will try to take more time to review all this more in depth in the coming days.

automenta:
agreed.  some of the changes will be obvious, but others can only be explained by my unorganized application of a sampling profiler and intuition :biohazard:  

IMO use of JDK8 parallel Stream's provides a cleaner code result but there isnt great control over the threads as they go into the common forkjoin pool.  something like the CountDownLatch at least is better than the Thread join wait loops i found.  in another project i've been using a hashed wheel timer feeding into a disruptor queue for dispatch to a fixed # of worker threads and that can give fine grain control over parallel cpu resources that i think is required here.  but if the current way works id still like to see metrics that indicate it.  when i look at the sampling profiler i am not sure the overhead of the threading isnt negligible

automenta:
example for the JDK8 Streams https://github.com/automenta/yacy_search_server/blob/21bb4093a00979e1fa8803af0a91bcaa7d986e84/source/net/yacy/peers/Network.java#L368

luccioman:
Some additional feedback :
- Base64Order.java
  - `(x > 0) ? 1 : (x < 0) ? -1 : 0;` is not strictly equivalent to `Integer.compare(x, 0);` as the contract of Integer.compare is to return "the value 0 if x == y; a value less than 0 if x < y; and a value greater than 0 if x > y". So if a JDK implementation of this function would not return -1 and 1 values, calling code would eventually break if relying on these specific values (and there is some in current YaCy code base)
  - `new byte[0];` is also not strictly equivalent to `ArrayUtils.EMPTY_BYTE_ARRAY`. If some calling code do some object references comparison it may eventually break. What's more, I guess this modification is for a supposed performance gain, but some [benchmarking](https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_caching_the_array) found on the subject does not look very convincing

- CollectionConfiguration.java
  - `for (int i = 0; i < t.length; i++) try {
		            t[i].join(10000);
		            if (t[i].isAlive()) t[i].interrupt();
		        } catch (InterruptedException e) {}` is not exactly equivalent to `latch.await();`, as with the await() version there is no limit on the waiting time
  - not sure of the expected benefit of `           Iterator<String> ii = collection1hosts.keys(true);
            while (ii.hasNext()) {
                String host = ii.next();` over `for (String host: collection1hosts.keyList(true)) {`

automenta:
thanks for checking

the integer compare came from intelliJ inspections and yes the semantics are different however i dont disagree with its recommendation to follow compareTo convention even if it means changing other code.

using a global empty byte array should perform better and use less memory than having to maintain N less references overall except if every instantiation of an allocated empty byte array happened to be completely eliminated by escape analysis.  however its probably negligible if it doesnt happen often.  im skeptical of that benchmark and its version - early access jdk9 from 2 years ago.  btw im using jdk12 built directly from openjdk.java.net mercurial and it seems to run yacy ok as does jdk9 and jdk10.  also JMC has a nice profiler http://openjdk.java.net/projects/jmc/
 
latch has a timed await method and its remaining countdown can be polled after fixed delay, it should still be an improvement over polling each individual thread.
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html#await(long,%20java.util.concurrent.TimeUnit)
another option is to use the latch so the last finishing thread upon detecting count=0 can call the postcondition.  then after the threads are submitted it can immediately return to callee without any polling.  

the iterator change was an attempt to avoid buffering a temporary list and instead stream directly from the source data structure, however the semantics between these is likely different if the underlying data structure changes during the iteration. even a concurrent data structure can not be expected to update the live iterator.  its a tradeoff either way depending on the collection size and timing

Orbiter:
This is a "mass change of everything", a very dangerous thing which - as @luccioman mentioned - cannot be reviewed. I tried and found out that there are a lot of "I just change the coding style" without functional changes.
Please make small pull requests, each pointing out which particular enhancement is made.

PR #220, opened 2018-08-20
Improve Dockerfile

pnhofmann:


luccioman:
Hi @pnhofmann , nice improvements, but one is debatable to my mind : "Download yacy as .zip instead of git clone". There is one good reason to clone from git : the Yacy exact version is calculated at build time based on information available in git history. If we build from zipped sources, the version (displayed notably in /Status.html) stays always at 1.921/9000, which is not very convenient and accurate when you check your YaCy node status...

pnhofmann:
Hi @luccioman , you are right, didn't notice that one. Just removed that commit.

luccioman:
Thanks @pnhofmann !

PR #238, opened 2018-09-25
[WIP] Spanish translation

ivanhercaz:
I began to translate YaCy some months ago, but I didn't commit any file nor progress. I have fork the current repository to be sure that I don't forget to translate some new string. This PR is work in progress, but I hope to translate everything soon.


luccioman:
Hi @ivanhercaz , that's great you initiated this for Spanish speaking users.
An advice to help you in this big task : regularly check your translated pages in a running YaCy peer to be sure nothing is broken in the translated pages. Maybe you already do that, otherwise you can have a look at commit eb94986f959024174b9d9bbc77dee98874be6c88 as an example of what code modifications are necessary to add a language in the list of YaCy interface available languages.

ivanhercaz:
Hi @luccioman!

Thank you for your comments. Yes, I am checking the translated pages to check there is nothing broken. I usually check it when translate a set of pages, like "Blacklist*_p.html".

About the commit eb94986, thank you very much! I had many doubts about it and I had to ask it, but now you give me the solution. In a near future I will make this change!

luccioman:
@ivanhercaz you can safely ignore the commented translations in the [language].lng files as you are starting a completely new translation. The various translators let them for various reasons, and some cleanup should probably be done.

About the translation editor: the entries that are displayed in it are the ones present in the master.lng.xlf file. Eventually, you may realize that entries are missing on some pages to generate a complete translation. Then you can either update the master.lng.xlf with new necessary entries and include these modifications in this pull request, or just manually add missing entries to your es.lng file. In the end your additional translations should appear in the translated page, so one should not forget that the translation editor is first of all here to help in the translation process, not to slow you down.

ivanhercaz:
All right @luccioman, I will ignore the commented translations and remove it from the es.lng to clean the file from unnecessary strings.

So the entries displayed are not from es.lng. The entries are from master.lng.xlf, the translated text from locales/es.lng and the empty strings with entries in English is because the entry is in the master.lng.xlf but it isn't translated in locales/es.lng and in the language file I copied (de.lng).

> so one should not forget that the translation editor is first of all here to help in the translation process, not to slow you down.

Of course! It is a great tool that we, all interested users in translate, should use. Translate directly from the source could be hard in some cases. I just use vim to check the both files, es.lng in locales folder and es.lng in DATA/LOCALE folder, to copy the strings from this last to the main es.lng

Thank you again!

luccioman:
Ok @ivanhercaz , no problem if you have no more time for now on this subject, the work you have already done is valuable, and most of other language translations are not complete.

Anyway, before merging, it would be great if you could take some time just to clean up remaining untranslated entries (that are still translated in German in your es.lng file for example).

luccioman:
@ivanhercaz your work is now available on the master branch.
Additionally, I just cleaned up untranslated entries and fixed a small issue with the "discover from index" button on the ConfigHeuristics_p.html page translation.

PR #239, opened 2018-09-25
Display correct time in Rejected URLs overview

otteresk:
Instead of displaying the current time, show date and time of entry's FailDate


luccioman:
Thanks @otteresk !

PR #240, opened 2018-09-30
Fix English Bookmarks.html

ivanhercaz:
Translating to Spanish I checked in the translator editor there is a string with a typo error. This PR fix the word ``serach`` to ``search``. I don't know if it is necessary to make more changes to solve it, but if there are more necessary changes, tell me and I will try to make it.

Regards,
Iván

### DIFF
--- a/htroot/Bookmarks.html
+++ b/htroot/Bookmarks.html
@@ -239,7 +239,7 @@ To see a list of all APIs, please visit the <a href="http://www.yacy-websuche.de
                     </table>
                     #(/autosearchrunning)#
                     <br />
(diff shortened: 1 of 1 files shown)


luccioman:
Thanks @ivanhercaz . The general idea when you modify some existing text from a YaCy html template is to check if it is also already present in the .lng or .xlf translation files, as you can see in last commit d42f079.

ivanhercaz:
All right @Luccioman! I will keep in mind your instructions and if I find another strings that needs to be modified, I will make the changes in all the pages you.

PR #258, opened 2018-12-05
Grammar change

agnelvishal:


luccioman:
Yes @agnelvishal that readme section is not very clear. But indeed the scripts call web interfaces and not the contrary. So I think a more proper fix can be : "They just call also the web interface pages."

PR #260, opened 2018-12-06
Update up.sh

agnelvishal:
IP addresses are not selected if "&ip=" is present


luccioman:
Thank you @agnelvishal .

PR #261, opened 2018-12-06
Update searchall1.sh

agnelvishal:
The up1.sh does not exist. up.sh does the work


luccioman:
Thank you again @agnelvishal !

PR #283, opened 2019-02-19
Update zh.lng

tangdou1:
more accurate.


luccioman:
Thank you @tangdou1 for your help.

PR #292, opened 2019-04-14
Fix a typo in french stopwords : "celà" => "cela"

zeroheure:


luccioman:
Thanks @zeroheure for the feedback. 
However please note that "celà" appears to be a non official but known alternative to "cela" (see notably the French WIktionary related [page](https://fr.wiktionary.org/wiki/cel%C3%A0). And indeed both spellings are in the [stopwords_fr.txt](https://github.com/apache/lucene-solr/blob/releases/lucene-solr/6.6.5/solr/example/files/conf/lang/stopwords_fr.txt#L173) file from the Solr 6.6.5 release. I upgraded YaCy copy of this file accordingly.

zeroheure:
I' didn't know that ! thanks for the explanation.

Issue #353, opened 2020-04-26
json-simple 1.1.1 is ancient


onlyjob:
With Android's `org.json.*` library it should be possible to replace 8 years old library `org.json.simple` (json-simple-1.1.1.jar) and unify on one JSON imlermentation.

Alternative this library can be upgraded to at least 2.3.0.

I've replaced `json-simple-1.1.1.jar` with Debian's `json-simple-2.3.0.jar` from `libjson-simple-java` package and everything seems to work well without any changes to the code whatsoever.

Orbiter:
this is a valid request but I would like to migrate to gradle first, see https://github.com/yacy/yacy_search_server/issues/377

thkoch2001:
#473

PR #356, opened 2020-05-05
Upgraded json-simple JAR to 2.3.0 (#353)

onlyjob:
Jar is taken from Debian's "libjson-simple-java" package 2.3.0-1.


Orbiter:
This is a complex, 27-files-changed pull request. That is not as "json-simple JAR to 2.3.0" says. Furthermore I did upgrade of some libraries already so we have a bunch of conflicts

onlyjob:
There is nothing complex here if you look at the commits. I've started upgrading libraries with trivial case of _json-simple_ then added few more trivial commits to upgrade few more libraries (and did not update pull request title). 27 files because some libraries have multiple .JAR files - that should come without saying. There is exactly one line removed for compatibility and everything else is upgraded .JAR files that makes this pull request _trivial_, not "complex".

I'd appreciate if you could check if some commits from this pull requests are still applicable.

PR #405, opened 2021-03-30
Improve HTML entities support

jfhs:
This PR has two HTML entities related improvements:
1. Switches from using hardcoded list of HTML entities that was missing some common entities and a lot of new HTML5 entities, with a loadable json file in `defaults` folder. Data itself is taken from https://github.com/mdevils/html-entities/blob/master/src/named-references.source.json and that project is MIT licensed, so should be fine to use it here.
I've run "builtin tests" from `CharacterCoding.main` and all looks good. Also verified that entity that I first noticed wasn't decoded during crawl was correctly decoded (&amp;mdash;).
**Edit:** I've tested some more, and adding all new Character -> Entity mappings causes issues for XML responses. Main problem seems to be that page handlers are shared between output formats (json, html, xml), but decide on encoding regardless of it. I.e. use `putHTML` or `putXML`, even if output is JSON. I assume that should've been actually done in template engine (which knows what format it's outputting), but changing that is out of scope and likely will cause more problems than it solves.
Reverted to using old tables for Character->Entity mapping.

2. Parses HTML entities for all attributes by default, instead of doing it whenever attributes were used. Forgetting to decode leads to incorrect data or some unfortunate loops, like one that can be observed on github page like this: https://github.com/trending/php&
If you check source code, there will be `<link rel="canonical" href="https://github.com/trending/php&amp;" data-pjax-transient>` and as it happens, href's for links were not decoded. So YaCy would see `https://github.com/trending/php&amp;` as a new link, fetches it and gets `<link rel="canonical" href="https://github.com/trending/php&amp;amp;" data-pjax-transient>`. As you can see, this will continue looping.
I see that decoding all attributes by default might be wasteful and reduce performance somewhat, but it's the simplest and most robust solution. If you think that's not acceptable another option is to do it lazily (subclass `Properties` and override `getProperty`), or carefully going over all `tag.opts.getProperty` calls and making sure they all decode content.


Orbiter:
excellent work!

PR #419, opened 2021-08-19
Increase timeout of IndexBrowser_p operations to 5 minutes

iansmirlis:
On hosts with large number of documents, display of subpaths or reload404 get timeout before they manage to complete, thus it might display an empty subpath or reload404 does not work properly.

I increased it to 5 minutes. I think that it's better to have a large timeout limit than not working at all without showing any timeout error. Even better, such kind of timeouts can be in configuration.

Any better ideas are welcome.


iansmirlis:
On a second thought this does not seem a good idea, it's better to fix only reload errors. I will revert

PR #420, opened 2021-08-19
In case of reload404, load only failed documents

iansmirlis:
Since it's my first commit to the project, it took me a while how to fix this without falling too much into it:

- The problem: In large indexes (>400GB), after a few days of crawling, you may end up with some finished hosts with a couple of million documents loaded well, but also with some failed pages.

Trying to Reload 404 errors etc, in order to reload pages that might have some temporary problem, is almost certain that the query will timeout before getting all documents, thus you might not be able to reload any error page at all.

- Solution: At first I changed TIMEOUT to something like 5 minutes, but it's ugly. So I ended up loading only documents that have set a failtype_s, which are expected to be an order of magnitude fewer than loading all host documents.

A breaking change of this is that when user clicks on reload errors she will see only failed page counts for this hosts. I also removed pending pages in order not to be more confusing.

- Test: I tested on a production environment with >25 million pages, seems to works well. I think there is some room for improvement on IndexBrowser_p pages, any ideas are welcome.

### DIFF
--- a/htroot/IndexBrowser_p.java
+++ b/htroot/IndexBrowser_p.java
@@ -370,6 +370,10 @@ public class IndexBrowser_p {
                         q.append(" AND ").append(CollectionSchema.url_paths_sxt.getSolrFieldName()).append(AbstractSolrConnector.CATCHALL_DTERM);
                     }
                 }
+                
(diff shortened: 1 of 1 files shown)


reger24 (approved):
looks good to me

PR #436, opened 2021-12-27
Improved the Image search page to have bigger thumbnails, use a more area for results and have a smaller sidebar.

ZeroCool940711:
This PR has some changes to the Image Search page, the goal was to make it easier to see the thumbnails as previously they were too small, for this I made three changes:
- Previously the sidebar was using 25% of the page width, I reduced it to be only 15% as that's more than enough for the sidebar.
- The area where the results are shown was changed to be 85%, it was 75% originally.
- Finally I increased the thumbnail size to be 256x256, previously it was 96x96 and the maximum I found was 128x128, I think this was the size for when you click on them, I changed both to be 256x256 but so far I still get bigger images when I click on the thumbnails so I believe it has no or little effect for the preview but the main thumbnails are way easier to see and more comfortable to the eye than before.

The result is this:
NewImageSearchPage
And this is what it looked like on the old/original version was this one:
OldImageSearchPage

So far the result seems pretty good, the images might look a bit blurry but I think this is because they are being scaled to fit the container for the thumbnail, I tried to fix it by using `image-rendering: pixelated;` and it improved a bit but as a result the images now look pixelated, even then it's better than a blurry image, this change is not on the PR as not everyone might like it. I also had to hard-code the `width` and `height` value on the `htroot/yacysearchitem.html` file as I was not able to find where the original values were taken from so I could change it on the right place, this should be fixed and should be done properly but for now it's not a big deal so could be left temporarily as hard-coded values.

### DIFF
--- a/htroot/env/base.css
+++ b/htroot/env/base.css
@@ -258,8 +258,8 @@ tt, *.tt {
 
 .thumbcontainer {
   margin: 2px;
-  width: 128px;
-  height: 128px; /* 96px thumbnail + some lines of text */
+  width: 256px;
+  height: 256px; /* 96px thumbnail + some lines of text */
   float: left;
   /* Cut non square images not rendered by YaCy ViewImage */
   overflow: hidden;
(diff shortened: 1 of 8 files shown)


Orbiter:
cool

Issue #36, opened 2016-01-02
HTTP/S Proxy support with Yacy Server


sudheesh001:
```
http_proxy = 
https_proxy =
```
might be the environment variables already present in the system, there should be a way to tell Yacy to use internet behind a proxy environment or respect the http_proxy settings.

maxolasersquad:
Hmmm, when I first setup my server the above was true. A few hours later it was showing itself to be running in senior mode, so I guess it's working fine behind the nginx proxy.

luccioman:
@maxolasersquad correct me if I am wrong, but as I understand it, @sudheesh001 was talking about proxying outgoing http connections from YaCy (acting as a client, when crawling for example), whereas you are describing your YaCy peer as running behind nginx acting as a reverse proxy/gateway (receiving incoming connections first and transmitting them to your Yacy peer).
This would not be exactly the same configuration issues...

maxolasersquad:
I think you are correct. When my server was running in junior mode I assumed it was because of the nginx proxy and went searching the issues. I originally thought this was it. Since my server began running in senior mode after a while its clear that yacy had no issue running behind an nginx server.

sudheesh001:
Absolutely @luccioman . In this issue, I was basically referring to the ability for the yacy clients to communicate when behind a proxy by enabling the http_proxy and https_proxy environment variables.

Thank you @maxolasersquad for taking your time to deploy the server behind an nginx proxy. Please feel free to share the nginx configuration as an update to the documentation for running Yacy peers for others in the community who might want to setup Yacy behind an Nginx proxy. 👍

ehehdada:
Hello! 

I have also installed YaCy 1.92 docker version listening at port 8090 behind of my nginx, public https port 443 (http port 80 redirects to https), and looking for the right way to configure YaCy. In `Use Case & Accounts` -> `Basic Configuration` -> point 4, whatever I put as port is reverted to 8090 after clicking on `Set Configuration` button. The SSL certificate is at the nginx, so I am connecting nginx and YaCy through http port.

If I change the ports at `System Administration` -> `Advanced Settings` then, after restart, I am unable to access YaCy and I even have to remove the volume and deploy it again (yes, I updated nginx conf to point port 80 in YaCy.)

Enabling `Transparent Proxy Access Settings` seems to have no effect.

[...]

PR #440, opened 2022-01-11
Add setting for public facing port

lfuelling:
Hi,

this PR regards #315 and [my comment](https://github.com/yacy/yacy_search_server/issues/36#issuecomment-1010074336) on #36.

It seems like YaCy already supports setting the public facing port and I just added the setting for this. 

It is very likely that I forgot something as I've never worked on YaCy before. If I have to do more, please let me know.

### DIFF
--- a/htroot/SettingsAck_p.java
+++ b/htroot/SettingsAck_p.java
@@ -219,7 +219,7 @@ public class SettingsAck_p {
                 if(pport < 65535 && pport >= 0) {
                     serverCore.usePublicPort = true;
                     sb.peers.mySeed().setPort(pport);
-                    env.setConfig(SwitchboardConstants.SERVER_PUBLIC_PORT, publicPort);
+                    env.setConfig(SwitchboardConstants.SERVER_PUBLICPORT, publicPort);
                 }
             } catch (NumberFormatException e) {
                 // noop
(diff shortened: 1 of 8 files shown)


pditzel:
Oh yes please, need this to.

htroot/SettingsAck_p.java
```diff
@@ -211,6 +211,19 @@ else if (!filter.equals("*")){
                 sb.peers.mySeed().setIP("");
                 env.setConfig(SwitchboardConstants.SERVER_STATICIP, "");
             }
+
+            // publicPort
+            String publicPort =  (post.get("publicPort")).trim();
+            try {
+                Integer pport = Integer.parseInt(publicPort);
+                if(pport < 65535 && pport >= 0) {
+                    serverCore.usePublicPort = true;
```
reger24, line 220: this "serverCoreusePublicPort"  does not exist (compile will fail)

acdapkus:
I am a bit unclear as to how to use this.

I would like to allow access from the public internet through my firewall to a "data exchange" peer port only. The web interface should only be accessible then via a different port, which is behind my reverse proxy accessible via a VPN only.

Currently I have the VPN and reverse proxy setup and connecting to the web application via port 8090. I would like the data exchange peer port to be port 8091.

Am I correct in thinking that the "publicPort" setting is this data exchange peer port?

PR #443, opened 2022-01-25
Idea project files by Gradle not ANT

Dalethium:
As per the comment by @reger24 on #442 the project should be built by gradle, however, the issue I was facing was that lines 36 to 43 of the gradle build file were commented making the htroot not possible to work with, this commit changes the idea files to use gradle and uncomments the htroot classpath entry from the gradle build file. 

I believe the htroot should be included in the IDE classpath by default as, otherwise, it makes the servlets impossible to work with.


reger24:
@Dalethium 
as I'm not a frequent IntelliJ user
- could you please check if the latest commit works on your side w/o your pre-configuration 
      -  means start with a fresh git clone (see test case https://github.com/yacy/yacy_search_server/commit/65034be1eef899d113a1ecd5ba43d4ddfbc98de1 )

fyi @Orbiter

reger24:
@Dalethium 
regarding the servlets classpath - if I use it as compile target -> compiler ends up with wrong package name (as servlets have no correct package description).  In IntelliJ htroot at least as directory seems to by available by standard in project tree.
Other IDE like NetBeans make it not available in project tree (but only in a separate tab - file tree)

I also would like if servlets could be treated as normal java source but IMHO before we not find a solution for the package name issue - I think we are not lucky.

Dalethium:
> @Dalethium regarding the servlets classpath - if I use it as compile target -> compiler ends up with wrong package name (as servlets have no correct package description). In IntelliJ htroot at least as directory seems to by available by standard in project tree. Other IDE like NetBeans make it not available in project tree (but only in a separate tab - file tree)
> 
> I also would like if servlets could be treated as normal java source but IMHO before we not find a solution for the package name issue - I think we are not lucky.

I believe my previous PR regarding the ant build handled that actually, or at least it seemed to. It was definitely at least working as far as I could tell. I am going to test https://github.com/yacy/yacy_search_server/commit/65034be1eef899d113a1ecd5ba43d4ddfbc98de1 now, one moment

Dalethium:
image
Using [that gradle script](https://github.com/yacy/yacy_search_server/commit/65034be1eef899d113a1ecd5ba43d4ddfbc98de1) leads to this project root -- it's worth noting that I just built the dist zip for the code using https://github.com/yacy/yacy_search_server/pull/443/commits/dfc7a23e603eb2bf2466d4ae69553e530191b1df and launched it on a windows server, it worked fine. The below image is of #443's project root, including htroot files with no class errors or issues with the IDE.
image

reger24:
thank's

reger24 (approved):
as per own testing and confirmed by Dalethium see comment https://github.com/yacy/yacy_search_server/pull/443#issuecomment-1020970196
it works out of the box w/o .idea config-dir existing (keeping main dear clean)

So @Orbiter can add this pull request if he/you wishes, if not it imho works too.

Regarding the inclusion of the currently commented out htroot as source set, my IDE works with it, IntelliJ too - wo why not,  ->  ! don't know about Eclipse !

Orbiter:
wow, I am an eclipse user maybe since 15 years but recently my eclipse becomes unusable because the editor is so slow. I am very happy that your combined contribution is now opening up the opportunity to have an alternative, thank you!

PR #533, opened 2022-10-31
Shifting from Limit Stack to LocalStack didn't work

pr0vieh:
because of:
NoticedURL, shift from GLOBAL to LOCAL:  double occurrence

since Global and Local share the HostHash source this must always occur!

Deactivate Check for Now!

### DIFF
--- a/source/net/yacy/crawler/HostBalancer.java
+++ b/source/net/yacy/crawler/HostBalancer.java
@@ -284,7 +284,9 @@ public class HostBalancer implements Balancer {
      */
     @Override
     public String push(final Request entry, final CrawlProfile profile, final RobotsTxt robots) throws IOException, SpaceExceededException {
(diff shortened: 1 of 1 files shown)


Orbiter:
> since Global and Local share the HostHash source this must always occur!

why and where do they share the hash?
Global and Local Stacks are different objects, the "has" property where these stacks are shared are in the NoticeURL class, not in the HostBalancer class.

pr0vieh:
Since all Global Stacked URLs for (remote Crawl) rejected at Shifting to Local Crawl with: 
NoticedURL, shift from GLOBAL to LOCAL: double occurrence

pr0vieh:
After this Fix i can help Crawl my Own "Remote Crawl" Stack without this Patch alle URLs will Removed without Crawl from my "Remote Crawl" Stack

PR #560, opened 2023-02-11
clean obsolete stuff from ant build.xml

thkoch2001:
- properties htdocsWWW, RDFaParser are not used anymore
- Don't know why I prefixed ivy target descriptions with '-->'
- targets 'all' and 'compile' just forward to 'compile-core'

The latter is a result of commit 1e1107c97c3822 (2022-10-02): "clean-up and new servlet method caching"

I came back to YaCy, wanted to compile and had to open build.xml to decide whether I wanted to call all, compile or compile-core.

The target name 'all' is very generic and does not indicate what 'all' actually means. I removed the target and changed the github workflows to use 'compile' instead.

### DIFF
--- a/.github/workflows/ant-build-pull-request.yaml
+++ b/.github/workflows/ant-build-pull-request.yaml
@@ -13,4 +13,4 @@ jobs:
         java-version: '8'
         distribution: 'temurin'
     - name: Build with ant
-      run: ant clean all dist
+      run: ant clean compile dist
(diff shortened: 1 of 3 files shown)


Orbiter:
this is a bit problematic because I believe there are too many scripts somewhere which may run "ant clean all" to build.
They would be broken. I would prefer to keep the "all" target

Orbiter:
see comment above

PR #576, opened 2023-05-02
Menu links

okybaca:
added a link to faq and legacy wiki to "help menu" as they were missing.
new documentation (FAQ) wasn't linked at all and wiki link is considered as a temporary help to users until everything is transfered to FAQ.


Orbiter:
One part of your changes had been part of another pull request, which I already merged. However, you are proposing links to the (old) wiki which is actually broken and I do not want to link it any more. Content of the wiki shall be moved to the home page as possible.

PR #580, opened 2023-06-18
added a restartYACY.sh script

okybaca:
added a restartYACY.sh script intended to be run periodicaly every day, to keep yacy instance fresh. 

it changed my instance uptime from several days (ending up in some memory consumption errors) to almost indefinite time

can be run by unix "cron" daemon, use crontab -e and add something like:
0 3 * * * /path/to/script/restartYACY.sh
to restart at 3 AM, every day

### DIFF
--- a/restartYACY.sh
+++ b/restartYACY.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env sh
+# Restart a running YaCy server through API call
+#
+# typical use is a periodic restart once daily using "cron" unix utility
+# 
+# use crontab -e
+# and something like:
+# 0 3 * * * /path/to/script/restartYACY.sh
+# to restart at 3 AM, every day
+#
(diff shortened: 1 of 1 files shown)


smokingwheels:
How much physical Memory is in your system?
Is your Yacy Java memory set to about 40% - 45% of total memory?

I found this in my travels.
[https://docs.cloudbees.com/docs/cloudbees-ci/latest/jvm-troubleshooting/#recommended-options](https://docs.cloudbees.com/docs/cloudbees-ci/latest/jvm-troubleshooting/#recommended-options)

okybaca:
Yes, I had plenty (16 GB), initialy I had about half of it reserved for YaCy, but every 1M crawled pages the memory footprint increased by ~1 GB and forced me to increase the limit, as [described here](https://community.searchlab.eu/t/topic/1088/9?u=okybaca).

PR #596, opened 2023-09-27
Added executable start stop commands in QB64

smokingwheels:
Saves typing ./startYACY.sh


joestr:
I don't really see the need for another pair of launchers.

This introduces another complexity in developing YaCy and managing the application.

smokingwheels:
@joestr
It's more convenient for the end user with a desktop OS.

Orbiter:
while I also see the comment from @joestr as valid, I cannot pull this at least for formal reasons because there are (not only large) undocumented binaries, executables with unknown origin or construction details.

PR #600, opened 2023-10-10
UI: modified link to Process Scheduler in left menu

okybaca:
Typical use of Process Scheduler is recalling the recent action. After entering Process Scheduler from left menu, sorting by date ascending mode is entered, user sees the very first actions (such as initial setup) and has to click Recording date to reverse the order. 

To ease this action I changed the default sort mode of Process Scheduler linked from left menu to "-date_recorded".


okybaca:
Thanks, Michael!

PR #607, opened 2023-10-29
replaced all the links to legacy legacy wiki to legacy wiki

okybaca:
since a lot of links leading to defunct yacy-websuche.de/wiki/ persisted in yacy, i replaced them (find-and-replace using sed) with wiki.yacy.net. 
i know, that wiki is going to be abandoned, but probably better to have legacy links than defunct ones


Orbiter:
ok good but sometime in the future we should remove those as well and point only to the home page.

okybaca:
yes, definitely!

PR #617, opened 2023-12-03
Add setting for DHT receive loadprereq insted of hardcoded load < 2.0

pr0vieh:
This helps to run full-fledged Yacy on servers as a side project


okybaca:
Thanks for a great patch, @pr0vieh!

PR #621, opened 2023-12-18
turkish translation tr.lng

okybaca:
turkish translation by @OFA54


Orbiter:
great work, thank you!

PR #651, opened 2024-07-09
removed yacy/ui as obsolete

okybaca:
'new ui' has started around 2007 and was never actually finished. present in the all the running instalations, poses a unnecessarily large attack surface.
removed `htroot/yacy/ui/` therefore.


Orbiter:
true 😢 sorry @apfelmaennchen 💟

PR #690, opened 2025-03-28
small fixes and fix all test

henschi:
**What does the PR**

- all tests can be compiled again
- fix a command line error in _Html2Image_, see test
- fix a cast error in _GenericXMLParser_, see test
- fix a NPE in _pptParser_, see test
- adapt _yacyVersionTest_ to the new behavior
- adapt _ParserTest_ to document data
- adapt _ooxmlParserTest_ to document data
- fix http/https links in _GenericXMLParserTest_
- adapt _WordTokenizerTest_ to the new behavior, but I think that “word,word” is not a word
- adapt _QueryGoalTest_ to the current behavior, but I think that additional space is not correct

As a result, all tests are now running.
(Ant unfortunately still reports two errors, as the test classes do not contain any test methods, only a main.)


okybaca:
> Cool, the CI check is green.  ;-)

Great! Good job!

henschi:
I add in class GenericXMLParser an additional fix in cast brackets.

PR #710, opened 2025-05-29
Maven proposal - happy discussion

henschi:
This is my actual Maven branch. I use it to compile Yacy and build a full Yacy release.

Useful Maven commands:

- mvn clean package
This makes a full compile and run all tests. The target jar can I used in a current Yacy installation.

- mvn -P release clean package
This creates a release package and run all tests. In RELEASE folder is then a release tar.gz

- mvn -P report clean verify -DskipTests
This creates an OWASP report and creates a SBOM.

The created Yacy release runs without any problems on my instance.
But a few dependencies are different to the Ant release.

Be careful with the branch, because I had to remove the Eclipse specific files!

What do you think?


okybaca:
henschi, that's great, thanks!

joestr:
This looks good!

3ele-projects:
Is there a foreseeable time or a decision whether Yacy will replace Maven with Ant? What exactly is stopping this M.R.?

(maybe there is the possibility to merge it into the repo as a separate branch)

okybaca:
It seems, it would be a great thing for developers, who need some debugging tools!

How far do you thing we are from completion?
What is neccessary to do, how could someone help?

PR #718, opened 2025-06-24
disable docker on forks

henschi:
Please disable Docker builds on forks, because forks can't upload the image.


henschi:
of course also pull requests

Issue #622, opened 2024-01-09
/date search is sorted by index time, not by date published


okybaca:
/date search uses the “date indexed” to sort out the results.  If I crawl a huge news site, all, even the really historical pages (NYtimes got archives dating back to 19th century, for example) are dated as “today”. 

Because I search the news, this is a huge use-stopper for me. 

Furthermore, in /date search results, some entries are dated even in the future, which is undesired behavior.

Would it be possible to do some heuristics on a real date published, probably using some combination of metadata?  
Other search engines do, somehow.  
Is it possible to switch /date operator to use time published, instead of the indexing date?

Some examples of date extraction from other software:

https://lyndon.codes/2019/05/03/extracting-date-times/
problem analysis + java

https://htmldate.readthedocs.io/en/latest/
python library

https://github.com/agnelvishal/newspaper/blob/master/newspaper/extractors.py
(python, line 192 onwards)

okybaca:
it's probably a question for @Orbiter : is there some field in the solr scheme, intended for storing of 'date published'? 
if not, how should be named? `published_date_dt`?

okybaca:
We discussed the topic with @orbiter at 38c3.

It seems that `last_modified` header doesn't reflect the date of publishing of information anymore, being more technical information about when the page was generated (maybe proxy and caching solutions are the reason for that).

I made a little research of the news sources from [news aggregator](https://eldar.cz/news). Most of the news sites publish the accurate date of publishing in [schema.org](https://schema.org/Article) Article metadata like:
`<span itemprop='datePublished' content='2025-01-26T20:21:00+01:00'>`, often using  JSON-LD.
schema.org metadata contains also a `dateModified` field, usualy, when the actual article was edited or something added. 

I see two possible solutions:
1) Either:  do a heuristic find of actual date published, prioritising schema.org `datePublished`, maybe using other leads, such as og `article:published_time` and maybe even url, as suggested [here](https://lyndon.codes/2019/05/03/extracting-date-times/). Date modified could be parsed as well, since it's present in both [schema.org](https://schema.org/dateModified) and [open graph](https://ogp.me/#type_article).
From my experience, `last_modified` header is sometimes far in the future by webpage-creator mistake. So the largest possible value should be current time.
I'd prefer this solution.
2) Or: do schema.org metadata parsing and store these in solr. Would  require more solr schema changes, but also could provide much more usable information (events, people, places, etc...)

[...]

PR #730, opened 2025-08-25
support for /date search to be sorted by date published #622

szarowski:
Support for /date search to be sorted by date published - https://github.com/yacy/yacy_search_server/issues/622

The set of rules for extracting date to fill last_modified in ContentScraperDateUtil.java have been carefully selected and tested by @okybaca. 

The getDate() method in ContentScraper.java has a new parameter Date lastModified due to passing the value of last_modified value from the HTTP header at the time of extracting it. That's why there are so many files are affected. The most of the calls pass null, just to fulfil the contract. It is used only in parseDocument() method in Switchboard.java when response has the lastModified HTTP header value. It can be applied later as one of the last use cases in the set of rules for extracting date to fill last_modified in getDate() method in ContentScraperDateUtil.java.

### DIFF
--- a/source/net/yacy/cora/date/CustomISO8601Formatter.java
+++ b/source/net/yacy/cora/date/CustomISO8601Formatter.java
@@ -1,0 +1,27 @@
+package net.yacy.cora.date;
+
+import java.util.Collection;
+import java.util.Set;
+import java.util.StringTokenizer;
+
+public class CustomISO8601Formatter extends ISO8601Formatter {
+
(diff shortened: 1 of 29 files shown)


okybaca:
tested thoroughly and works great! 
detects last modification of page on most of news sites, using heuristics and various metadata, and writes correctly into last_modified solr field.
great work @szarowski, thanks a lot for your contribution!

Orbiter:
great! @szarowski @okybaca thank you for your work!

PR #754, opened 2026-01-21
Add configurable DHT error blocking with retry window

pr0vieh:
### Summary

- Add proactive DHT error URL rejection for both URL and RWI transfers, with configurable opt-out.
- Reject known-broken URLs/RWIs on receive when httpstatus_i != 200 or failreason_s set.
- Permanent errors (404, 410, -1 DNS/network) are never redistributed; temporary errors (5xx/timeouts) get a retry after N days (default 30).
- Sender learns rejected hashes via errorURL feedback and marks them locally.
- Feature is opt-out and fully configurable (toggle, retry days, permanent status list) with UI controls in IndexFederated_p.

### Why accept

- Less useless DHT traffic: broken hashes are stopped at ingress instead of being forwarded.
- Cleans the network: known dead/NXDOMAIN targets are removed from circulation.
- Prevents crawl loops: peers avoid re-crawling URLs already known to fail.
- Faster convergence: temporary failures get a timed backoff; permanent ones are pruned immediately.
- Safe defaults, but overridable; transparent UI controls; symmetric handling for URL and RWI paths.

### Config defaults

indexReceiveBlockErrors=true
indexReceiveBlockErrors.retryAfterDays=30
indexReceiveBlockErrors.permanentStatus=404,410,-1 (-1 = DNS/network)

### DHT crawl loop schema
```
[1] Peer B holds old URLs 
    |
    v
[2] B DHT-pushes those hashes -> A
    |
    v
[3] A receives, sees httpstatus != 200
    - marks as failed, schedules retry if temporary
    - sends errorURL feedback to B
    |
    v
[4] B receives errorURL feedback
    - marks same hashes as failed
    - stops pushing them to all
    |
    v
[5] Later, if A cleans its local fail list:
    - B still won’t re-send; hashes remain blocked on B
```

### DIFF
--- a/defaults/yacy.init
+++ b/defaults/yacy.init
@@ -295,7 +295,7 @@ proxyCacheSize = 4096
 
 # The compression level for cached content
 # Supported values ranging from 0 - no compression (lower CPU, higher disk usage), to 9 - best compression (higher CPU, lower disk use)
-proxyCache.compressionLevel = 6
+proxyCache.compressionLevel = 9
 
 # Timeout value (in milliseconds) for acquiring a synchronization lock on getContent/store Cache operations
 # When timeout occurs, loader should fall back to regular remote resource loading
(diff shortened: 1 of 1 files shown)


okybaca:
Just thinking aloud, prove me wrong:
from a security perspective: isn't it a way in which a mailicious actor can "remotely delete" other peers RWI indexes?
is it safeguarded against that?

PR #757, opened 2026-01-23
Persist recrawl job to restart on peer boot

### LINKED ISSUE #745: Small recrawler bugs & enhancements
- [x] recrawl is stopped after restart
   - it would be nice, if the recrawl job would continue even after restart (which is in many cases daily)
   - i'd set up even an periodic recrawl, but recrawling doesn't show in Process Scheduler - how to show actions there?
   - **solved by #757** 

- [ ] collection of recrawled pages is changed to `robot_recrawlJob` instead of keeping the collection name  
  - which is a bit against concept of collections and exporting data in packs - once recrawled, assigned collection name is lost 

- [ ] recrawl progress is not shown in `crawler monitor`, PPM is shown
  - is there a reason for that?

pr0vieh:
fix #745

### DIFF
--- a/source/net/yacy/htroot/IndexReIndexMonitor_p.java
+++ b/source/net/yacy/htroot/IndexReIndexMonitor_p.java
@@ -165,6 +165,18 @@ public class IndexReIndexMonitor_p {
             			}
             			sb.tables.recordAPICall(post, IndexReIndexMonitor_p.SERVLET_NAME, WorkTables.TABLE_API_TYPE_CRAWLER,
             					"Recrawl documents matching selection query : " + recrawlQuery);
+						/* Ensure the recrawl job restarts on peer reboot */
+						final Row recrawlCall = WorkTables.selectLastExecutedApiCall(IndexReIndexMonitor_p.SERVLET_NAME, post, sb);
+						if (recrawlCall != null) {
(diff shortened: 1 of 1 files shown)


okybaca:
Thanks a lot for the patch, @pr0vieh!
Feels great, that someone actually fixes the issues.
I appreciate that a lot!
<3

okybaca:
tested "persist recrawl job to restart on peer boot" and works great, in my
case. thanks, @pr0vieh!

okybaca:
The problem with the patch is, that it (re-)starts the reCrawl job, even
though it was previously terminated.  Which means, once recrawl job is once
run (and terminated later, with a clean queue and status: terminated), it's
started on every restart, resulting in the unwanted run.

Log:
`
APICALL * /IndexReIndexMonitor_p.html?transactionToken=&deleteOnRecrawl=on&recrawlnow=start%20recrawl%20job%20now&recrawlquerytext=fresh_date_dt:%5B*%20TO%20NOW/DAY-1DAY%5D&setup=recrawljob&includefailedurls=on&
`
during fresh start. 


Expected behavior is conditioned: when recrawl runs before restart, it's
started again after.  If it doesn't run before, it's not started.

Issue #625, opened 2024-01-15
Large blacklists freeze the UI for importing a blacklist


Avamander:
It would be really nice if larger blacklists (around 10 000 entries) were paginated in the blacklist import view. Even larger blacklists (imported malicious domain lists) cause the page to not render and crash.

Avamander:
At some point the page is just too large for the browser to render it properly.

Avamander:
I tried that, but it still pollutes index view unfortunately. (And there's no easy "blocklist and delete from index" button there)

Avamander:
> In the Index Administration menu there's an option to delete the url from the index and in filters you can blocklist it through regex (not always work tho).

Pretty sure it's the opposite, there's a button to blacklist a domain in index search/viewer but you have to delete using regex under index management. There's no button in the index explorer (that shows an overview of all the domains).

smokingwheels:
I have found if you append the url.default.black file Yacy accepts it.

📝 Explanation: Extending url.default.black Using a Simple QB64 Script

If you want to append a large list of hosts into YaCy’s url.default.black using a script, the following QB64 snippet works reliably:
```
Open "allhosts" For Input As #1
Open "url.default.black" For Output As #5

Do While Not EOF(1)
    Input #1, a$
    Print #5, a$ + "/*"
Loop

Close #1: Close #5
```
✔ What this does

Reads each line from allhosts a blocklist.

Appends /* to the end of each entry
→ This turns each hostname into a prefix wildcard rule, which YaCy accepts

Writes the expanded rule into url.default.black
```
Example input line:
example.com

Becomes:
example.com/*
```
This ensures that all subpaths under that hostname are blocked by YaCy’s blacklist mechanism.

✔ Why this works with YaCy

YaCy’s URL blacklist supports:
exact matches
prefix rules (host/*)
full wildcard patterns
Appending /* makes the rule more comprehensive and avoids missing URLs with trailing paths.

✔ Notes for developers
This script replaces the target file (For Output truncates).
To append instead of overwrite, use For Append As #5.
YaCy reads blacklist files at startup; after modifying, restart the node or trigger a reload.
For very large lists, using url.blacklist.pattern.load or merging via shell (cat) is also valid.

✔ Minimal shell version (if you want a one-liner)
```
sed 's/$/\/\*/' allhosts > url.default.black
```

Same effect, much faster for large lists.

PR #769, opened 2026-04-20
fix: run blacklist import in background thread to avoid UI freeze (#625)

songproducer:
The add operation iterated over all selected items, parsed them, and wrote to disk inline on the servlet thread. For large blacklists this blocked the whole request. Now: read all item strings from post before returning (thread safety), hand off the parse+write work to a daemon thread, and redirect immediately. Failures are logged via ConcurrentLog.warn instead of silently dropped.


songproducer:
BTW I used Hermes agent and Claude code Sonnet for these bug fixes

Issue #724, opened 2025-08-01
RWI: indexDistribution.minChunkSize does nothing


okybaca:
My local RWI index grows a lot while crawling, but it's distribution to other peers is really slow and it doesn't "dissolve" at all. 

Usually, just a small amount of urls/term is transfered in a single batch:

`Index transfer of 7 references for terms [ XXX ..] and 13 URLs to peer ...`

occasionally, up to 1000 references are sent. 

I would like to maximise the term/urls in a single transfer, to speed-up the process a bit. 

I found a [configuration option](https://yacy.net/operation/yacy_conf/#index-distribution-and-transfer) `indexDistribution.minChunkSize = 10` for that, in the configuration file, but it does nothing, nor it's referenced anywhere in functional code. 

My suggestion is to improve the code (probably in `transferRWI.java` or `Transmission.java`), so it reflects the `indexDistribution.minChunkSize` setting (`indexDistribution.maxChunkSize` as well), allowing optimisation of RWI transfer.

As I can hear from disk activity, that this is quite io-heavy operation, so the option should remain configurable by user, to fine-tune the local instance.

PR #771, opened 2026-04-20
fix: wire indexDistribution.minChunkSize and maxChunkSize config to R…

songproducer:
…WI transfer batch (#724)

INDEX_DIST_CHUNK_SIZE_MIN and _MAX constants were defined but never read. dhtMaxReferenceCount was set only from startChunkSize with no bounds enforcement. Clamp the initial value to [minChunkSize, maxChunkSize] so both config keys actually control the DHT reference count range.


songproducer:
BTW I used Hermes agent and Claude code Sonnet for these bug fixes

PR #780, opened 2026-04-25
feat: cap results per domain to 2 to improve result diversity

songproducer:
Without this, a single high-scoring domain (e.g. Wikipedia, Reddit) can fill most of a result page. shouldEmitCandidate now tracks per-hosthash emission counts and drops results once a domain has contributed MAX_RESULTS_PER_DOMAIN (2) entries. The cap is bypassed for site:-scoped queries where the user intentionally wants results from one domain.


Orbiter:
Maybe there are examples that this is useful, but reducing results to two hits per domain is a complete new behavior that is not transparent. If ranking demands more hits per domain, then it is so. For my taste this is too much moderation.

PR #785, opened 2026-04-25
feat: show 'N more from domain.com' link when domain cap hides results

songproducer:
When the domain diversity cap (MAX_RESULTS_PER_DOMAIN=2) drops results for a domain, SearchEvent now tracks the skipped count per hosthash. Each rendered result exposes this count to the template, which appends a clickable link 'N more from domain.com' that launches a site: search to recover the hidden results. The link only appears when at least one result was hidden.


Orbiter:
this requires https://github.com/yacy/yacy_search_server/pull/780 and in some way admits that the moderation may be too strict. It creates complexity. I recommend that search result moderation should be done on custom search interfaces that simply use the YaCy search api.

Issue #794, opened 2026-05-31
Issues with using YACY_ environment variables for camelCase settings


mmeier86:
Hello YaCy team,

I've been setting up my new instance using the `YACY_` environment variables, and I've found that while it works for some settings like `port` or `upnp.enabled`, it does not work for others, like `staticIP` or `browserPopUpTrigger`. The environment variables for the latter to just be ignored, and my `yacy.init` still contains `browserPopUpTrigger=true`, and my startup logs still show the log lines where it tries to launch the browser.

Digging through the code, I think I might have found the problematic place:
https://github.com/yacy/yacy_search_server/blob/6d8c08b34aca934927665eb90d3238934779c8d6/source/net/yacy/server/serverSwitch.java#L143-L149

I think the problem here is the `toLowerCase` here in line 146. Take for example the `YACY_BROWSERPOPUPTRIGGER` variable. The `key` value would end up as `browserpopuptrigger`, and I assume that the `configProps.containsKey` method does a case sensitive comparison and thus won't find the key, leading to the environment variable not being used.

For this one I'm afraid I wasn't able to come up with a good solution I could propose.

mmeier86:
I should have thought about this one a bit longer than two minutes. The solution would be to also check the key lower cased in the comparison, not just the env variable's name. I will provide a PR for this one during the weekend.

mmeier86:
I have created #797 with a fix.

PR #797, opened 2026-06-05
Fix env vars for camelCase settings

mmeier86:
Hi YaCy team,

This PR relates to #794 

While setting up YaCy with the Docker image in my k8s cluster, I found that environment variables only work for settings like `server.https`, which consist of all lower letters (and dots) only. But for settings like `browserPopUpTrigger`, the environment variable `YACY_BROWSERPOPUPTRIGGER` would be ignored.

It turns out that that that's because during setup, the environment variable is changed to all lower case to find the related config key, and that doesn't work for camelCase keys, as they aren't found.

This PR fixes the issue. As we can't derive camelCase names from the all-caps names of the environment variables, I decided to rewrite the loops a bit. I'm now looping over all entries in the `configProps` map, and check for each entry whether there is a corresponding env variable or system property. To not iterate over all configs twice, once for env variables and once for system properties, I combined them into the same loop. The order of precedence, with env variables overwriting system properties, is preserved.

One question I've got: I will admit that I haven't worked with Java applications in a very long time, but is checking the System properties actually useful here? Are those generally used for settings when deploying an app?

I've done some local testing with this commit, checking whether `camelCase` settings now work, and whether `all.lower.cased` settings still work, and both succeeded.

### DIFF
--- a/source/net/yacy/server/serverSwitch.java
+++ b/source/net/yacy/server/serverSwitch.java
@@ -133,19 +133,12 @@ public class serverSwitch {
         // Read system properties and set all variables that have a prefix "yacy.".
         // This will make it possible that settings can be overwritten with environment variables.
         // Do this i.e. with "export YACY_PORT=8091 && ./startYACY.sh"
-        for (final Map.Entry<Object, Object> entry: System.getProperties().entrySet()) {
-            final String yacykey = (String) entry.getKey();
-            if (yacykey.startsWith("YACY_")) {
-                key = yacykey.substring(5).toLowerCase().replace('_', '.');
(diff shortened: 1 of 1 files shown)


zouzou123gen:
Looks great, do you know if this still works on 1.942?
I hit the same issue, applied your patch, but it seemed to have no effect (or even worse, think it stopped dotted settings from getting through as well).

I ended up doing seds like in https://github.com/yacy/yacy_search_server/blob/c2421aeeb91489169d3f87b0116b75141e5d1c8d/docker/Dockerfile#L28-L31 but env vars would feel much nicer.

PR #831, opened 2026-09-20
Fix malformed query history records and date range boundaries

blinggnome:
## Problem

`AccessTracker.addToDump()` writes embedded CR/LF characters from search queries
directly into `queries.log`. The reader treats physical lines as records, but
only checks their length and the space at offset 14 before parsing a count.
Continuation text can therefore raise an uncaught `NumberFormatException` and
break the local-search history page (`AccessTracker_p.html?page=2`). The timeline
API uses the same reader.

A synthetic example of the old writer's output is:

```text
20260101000000 1 sq synthetic-start
abcdefghijklmn xx sq synthetic-continuation
20260101000100 1 qs synthetic-next
```

The second line passes the old header guard, and `Integer.parseInt("xx")` runs
outside its exception handler. Depending on where date seeking lands, the same
line can instead cause a date-parsing IOException and an empty history result.
The regression test inserts this continuation among 64 valid records to
exercise the exception path, and at every insertion position to exercise seeking.
All examples and tests use invented queries, not user search history.

Another malformed entry previously accepted a count of **72**:

```text
20260101000000 x qs synthetic-invalid-count
```

The one-character shortcut calculated `'x' - '0'` without checking for a digit.

The range search also returned the preceding record instead of the first record
at or after a boundary. It omitted the final record when `to` was past EOF,
returned nothing for a one-record file, and lost records with equal timestamps.
For example, a range from 00:10:30 to 00:20:30 over minute-spaced records returned
00:10 through 00:19 rather than 00:11 through 00:20.

## Change

- Normalize CRLF, CR and LF to spaces in the logged copy only. Search execution
  still receives the original query.
- Use shared strict timestamp/count validation during both seeking and record
  parsing. Skip malformed physical lines and continue reading valid records.
- Correct inclusive `from` / exclusive `to` boundaries, duplicate timestamps,
  empty files, trailing malformed text and EOF handling.
- Retain date seeking and buffered UTF-8 reads; bound the selected stream instead
  of allocating a byte array for the entire selected range.
- Preserve existing files and the five-digit count limit. This does not attempt
  to reconstruct legacy multiline query text or change the existing assumption
  that searchable log timestamps are chronologically ordered.

Only `AccessTracker.java` and its regression test are changed.

## Verification

- All **11 new regression tests fail against the original implementation** and
  pass with the fix, including the exact `NumberFormatException` reproduction.
- **19 focused tests pass** on both the development checkout and an isolated
  checkout based on upstream `e171a4a0e`: AccessTracker, QueryParams and
  GenericFormatter tests.
  The PR uses an older shared ancestor to avoid carrying workflow changes into
  the fork; its merge tree with `e171a4a0e` is byte-for-byte identical to the
  tested checkout.
- Dev HTTP test: submitted a synthetic multiline query through the Solr API and
  flushed the log through the timeline API. Before the fix, multiline history
  made statistics disappear. After deploying the fix with that history intact,
  the history page and timeline API returned HTTP 200 with useful history, and
  a new multiline query was persisted as one physical line. Local search also
  returned HTTP 200.
- A synthetic log with one million valid records plus malformed continuations
  (about 47 MB) returned the correct 3,600-record and 100,000-record windows in
  approximately 62 ms and 117 ms respectively with a 256 MB heap. These are local
  observations, not a performance guarantee.

`ant compileTest` passed on the dev checkout. Upstream application compilation
(`ant compile`) passed. On the upstream base, the full `compileTest` target omits
generated `solr9-bridge-*.jar` files from its test classpath and fails in existing
Solr tests. This PR does not modify that unrelated build issue. The focused
tests were compiled and run with the complete runtime classpath:

```sh
javac --release 11 -cp 'build/classes/java/main:lib/*:libt/*' -d test/java \
  test/java/net/yacy/search/query/AccessTrackerTest.java \
  test/java/net/yacy/search/query/QueryParamsTest.java \
  test/java/net/yacy/cora/date/GenericFormatterTest.java
java -cp 'test/java:build/classes/java/main:lib/*:libt/*' org.junit.runner.JUnitCore \
  net.yacy.search.query.AccessTrackerTest \
  net.yacy.search.query.QueryParamsTest \
  net.yacy.cora.date.GenericFormatterTest
```


blinggnome:
## Production rollout validation (2026-09-20)

We have now tested this fix against existing production query histories, including
nodes actively exhibiting the reported Local Search Log failure.

| Check | Result |
| --- | --- |
| YaCy instances verified | 83 |
| Updated during this rollout | 82; one previously patched canary was already healthy and was not restarted |
| Confirmed history-page parser failures before the update | 27 instances returned authenticated HTTP 500 |
| Those same failing pages after the update | All 27 returned HTTP 200 |
| Local Search Log after rollout | All 83 instances passed authenticated checks |
| Other application checks | Access-tracker page 1, local search and Solr checks passed on all 83 |
| Existing query history preserved | Approximately 8.14 GB across the 82 updated instances, verified unchanged; the earlier canary was checked separately |

The failing endpoint was `AccessTracker_p.html?page=2`. Before/after checks used
authenticated requests to the actual YaCy servlet, rather than treating an
unauthenticated login page or a successful root-page response as proof of repair.
The existing query histories were not deleted, truncated or rewritten to clear
the error. Only the patch and optional matching source were installed, followed
by a YaCy service restart; no operating-system reboot, index cleanup, settings,
blacklist or dependency changes were part of this rollout.

No relevant parser or linkage errors were found in the retained post-start
application logs examined by the rollout verifier. This was a bounded check of
available rotated logs, not a claim about unlimited historical log coverage.

These results provide real-world confirmation beyond the synthetic regression
tests: the failing pages recovered while their original histories remained
intact. They describe post-deployment application checks, not an overnight soak
test or a fleet-wide external routing/DNS validation.

This summary intentionally contains only aggregate results. No node identifiers,
hostnames, IP addresses, local filesystem paths, credentials, runtime fingerprints,
raw HTTP bodies, query text or private report attachments are included.

